PR #1, opened 2024-11-03
Repo Restructure

SergeBakharev:
- Separate project files and application code.
- Moved to using Poetry for dependency package management and version locking
- Wrote a pyinstaller build script, integrated with poetry
- Wrote a LICENSE and NOTICE file
- Added documentation to README describing how to locally develop and run the pyinstaller packaging
- Removed simfang.ttf, it's not referenced in any files and in my quick local testing the application continued to have Chinese language support correctly without it. This saves 10MB and came from a questionable source.

I was tempted to further separate the assets from the code, but I think it's nice for iterative development purposes to be able to run the application easily without messing around. Maybe in future, I'll write some kind of manifest file that documents which files should be packed in via PyInstaller, and which files should be loose and just bundled together.

I haven't tried to isolate bCNC and XMODEM into Poetry managed dependencies, I have strong suspicions that the files have been modified from source, and using new versions might have some rather unexpected implications.

### DIFF
--- a/carveracontroller/GcodeViewer.py
+++ b/carveracontroller/GcodeViewer.py
@@ -737,9 +737,9 @@ class GCodeViewer(Widget):
             glsl2 = 'simple_apk.glsl'
             glsl3 = 'axis_helper_apk.glsl'
         if not os.path.exists(glsl1):
-            glsl1 = os.path.join(os.path.dirname(sys.executable), glsl1)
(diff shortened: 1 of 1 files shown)


SergeBakharev:
[Release Carvera controller as open source so that the community can help add features/beta test](https://trello.com/c/6GBqzJ5G/48-release-carvera-controller-as-open-source-so-that-the-community-can-help-add-features-beta-test)

PR #49, opened 2024-11-10
Reversed a axis and feed rate scaling range fixes

SergeBakharev:
- Fixes the A axis rotation in the 3d viewer
- Increase the feed rate scaling range from 50-200 to 10-300. The stepping is still 10%

### DIFF
--- a/carveracontroller/CNC.py
+++ b/carveracontroller/CNC.py
@@ -292,7 +292,7 @@ class CNC:
 
             elif c == "A":
                 self.has_4axis = True
-                self.aval = value * self.unit
+                self.aval = value * self.unit * -1  # Correct rotation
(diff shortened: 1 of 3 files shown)


carveracontroller/CNC.py
```diff
@@ -292,7 +292,7 @@ def motionStart(self, cmds):
 
             elif c == "A":
                 self.has_4axis = True
-                self.aval = value * self.unit
+                self.aval = value * self.unit * -1  # Correct rotation
```
SergeBakharev, line 295:
```suggestion
                self.aval = value * self.unit * -1  # Right Hand Rule rotation. A+ movement when looking at rotary jaws is CCW rotation
```

PR #74, opened 2024-11-15
Create update_translations.py

faecorrigan:
a quick method of updating the .po files for localization. Note that these files will need manually translated. This should be run every release.

### DIFF
--- a/carveracontroller/update_translations.py
+++ b/carveracontroller/update_translations.py
@@ -1,0 +1,55 @@
+####
+#this program builds the translation files automatically from the main python programs
+#it requires gettext https://www.gnu.org/software/gettext/
+#on windows this requires installing https://gnuwin32.sourceforge.net/packages/gettext.htm
+
+#to add a new language, update the LANGS array in main.py with the two letter code and the display name. 
+
+#run with python update_translations.py while in the carveracontroller directory
+
+
+####
+
+
+import subprocess
+import os
+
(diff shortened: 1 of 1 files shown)


SergeBakharev:
Nice work with this one. Now we just how to decide how we will get the translations done.

Since it's used as part of the development of the application, the script should be moved from carveracontroller/update_translations.py to scripts/update_translations.py and file references updated.

carveracontroller/update_translations.py
```diff
@@ -0,0 +1,55 @@
...
+subprocess.run(["xgettext", "-d", "messages", "-o", pot_file, "--from-code=UTF-8"] + python_files)
+print(f"Generated .pot file from Python files: {pot_file}")
+
+# Process .kv files separately with --language=Python
+for kv_file in kv_files:
+    subprocess.run(["xgettext", "-j", "-d", "messages", "-o", pot_file, "--from-code=UTF-8", "--language=Python", kv_file])
+    print(f"Appended .pot file with entries from {kv_file}")
+
+# List of languages for .po files
+languages = ["de", "en", "es", "fr", "it", "ja", "ko", "pt", "zh-CN", "zh-TW"]
```
SergeBakharev, line 32: Let's remove all but the currently supported languages. Both from the script and also from the carveracontroller/langs/ folder. 

```suggestion
languages = ["en","zh-CN"]
```

PR #77, opened 2024-11-15
Pr bugreport fc241115

faecorrigan:
opens the issues pages for the firmware and controller githubs. Found in the same menu as settings

The icon is being used from open iconic and is under the MIT license. https://github.com/iconic/open-iconic

image


SergeBakharev:
Very nice. Do you think we will use more icons from this project?
If so I think the file pathing of `carveracontroller/data/open_iconic_UI/<icon>` makes sense, otherwise lets just put it into `carveracontroller/data/`

SergeBakharev:
Implements #78

PR #145, opened 2025-02-24
Custom Background Images For File Setup

faecorrigan:
a dropdown for the File Setup screen that grabs all the image files in a directory and puts them in a dropdown. This allows the user to see where the bolt holes are on the machine to avoid them when probing, or load in their own custom image with a vice silhouette for instance

If the dropdown is set to none it defaults back to stock behavior


carveracontroller/main.py
```diff
@@ -118,7 +118,7 @@ def switch_lang(self, lang):
 from kivy.lang import Builder
 from .addons.tooltips.Tooltips import Tooltip,ToolTipButton,ToolTipDropDown 
 
-
+PLAY_FILE_IMAGE_DIR = os.path.join(os.path.dirname(__file__), 'data/play_file_image_backgrounds')
```
SergeBakharev, line 121: Variables defined outside of methods are not initialized when the app is run via the pyinstaller packaging (since the entrypoint is main()).

Please define the variable in `load_constants()` instead.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 [Unreleased]
 - Enhancement: The ability to reduce the size of the autolevel probe area. This enables the ability to avoid probing the stock where there might be obstacles preventing probing.
 -Enhancement: tooltip support. See https://github.com/Carvera-Community/Carvera_Controller/pull/143 for more information
+-Enhancement: added custom background images in a dropdown for the start file/probe screen to show bolt hole positions. The user can create their own by replicating one of the ones in the existing foler
```
SergeBakharev, line 4: formating/spelling:
```suggestion
- Enhancement: Added custom background images in a dropdown for the start file/probe screen to show bolt hole positions. The user can create their own by replicating one of the ones in the existing folder
```

carveracontroller/data/play_file_image_backgrounds/C1 Bolt Holes.png
SergeBakharev, line 1: The holes on the far right are cut off, is that intentional?

faecorrigan: that is the way makera set the screen up, so the cutoff is intentional. It could be changed with a larger modification to the codebase on a later date

carveracontroller/main.py
faecorrigan: fixed with new commit

SergeBakharev:
As discussed on discord the `data/` folder is a read-only asset, so the user can't add custom background images there. Making it read/write is challenging because it would be not platform agnostic after packaging.

I suggest adding a controller config variable for the user to provide an path for custom backgrounds and scanning that in addition to the embedded `data` folder. If you want to just merge this in for now without making that change just remove references to the user being able to make their own.

carveracontroller/data/play_file_image_backgrounds/C1 Bolt Holes.png
faecorrigan: i wonder if things might be a bit weird with the air due to the worksize being different/causing a different zoom factor. Something to test. If so, it would be easy enough to modify the air image to fit

faecorrigan:
@SergeBakharev I created a new commit with updates for a custom folder location in the controller settings and updated the tooltip to match. When you select the new background button, it opens both the directories for the existing background images as well as the set custom background. This probably won't work on IOS

There are a couple of errors with the tooltip creation function that I will be looking into

PR #149, opened 2025-02-28
Polish config run screen

SergeBakharev:
- Fix crashes related to unset custom background images config
- Scope custom background images variable to CoordPopup instead of global
- Hide autoleveling offset text if not set
- Remove new background button from config+run screen


faecorrigan (commented):
Looks good to me, and tests well on my machine. I will do some investigation about how adding custom images might be implemented in the future. At a bare minimum the info will be available in a release video

PR #154, opened 2025-03-02
Upgrade code to latest Carvera official controller 0.9.10

zittix:
This is a work in progress CR to update the code to controller 0.9.10 in order to support Carvera Air and other new fixes.


SergeBakharev:
Need to move across my comments from #154 .

I suggest waiting a week or two, I think we heard that Makera plans on releasing a new Controller version very soon.

PR #155, opened 2025-03-06
Reduce tooltip delay from 2s to 0.5s

pzich:
Not sure if there is any opposition to this, but I'm finding the current 2 second delay before showing the tooltips to be rather slow, especially if you're trying to hunt for a specific option or browse what is available. This reduces it to 0.5s, although an even shorter (or possibly no) delay might be better? Thoughts?

## Before
https://github.com/user-attachments/assets/ae321169-138c-467a-b81e-2c35c67e3e4c

## After
https://github.com/user-attachments/assets/13623b9f-7c1a-4ff6-9e20-578923a62826

### DIFF
--- a/carveracontroller/addons/tooltips/Tooltips.py
+++ b/carveracontroller/addons/tooltips/Tooltips.py
@@ -27,7 +27,7 @@ class ToolTipButton(Button):
     tooltip_txt = StringProperty('')
     tooltip_cls = ObjectProperty(Tooltip)
     tooltip_image = StringProperty('')
-    tooltip_delay = 2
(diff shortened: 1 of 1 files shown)


SergeBakharev:
This feature is modelled to be similar to Fusion 360. In Fusion the delay is 1s, but is user configurable: https://www.autodesk.com/blogs/autocad/autocad-tooltips-time-tuesday-tips-lynn/

I suggest we match, and potentially implement similar config

SergeBakharev:
Merging this change as-is for now, I see no harm in having the popup at 0.5s. If someone complains, I can help make the delay configurable

PR #163, opened 2025-03-23
Adding Probing Control Screen to project

faecorrigan:
image

Adds GUI probing control for outside corners, inside corners, single axis probes with double tap, bosses, bores, and angles

Before pushing to main it would be good to get a quick sanity check on all the operations to make sure they do what is intended. I need to walk away from this for today and will do a personal sanity check on monday march 23

Todo for the next release:
fix a small issue with tooltips in boss/bores showing up when the confirm screen is active
add calibrate probe tip diameter and calibrate machine positions screens. These require graphics
make a few more graphics for tooltips
add information on where to start each probing operation to each screen
add help info buttons that lead to the community youtube videos on each probing operation that will be produced
add some form of warning to make sure the tool is a probe
add warnings/overrides for community vs stock firmware
consider changing the preview and run screen to a text input field so the user can make changes in that screen/copy/paste


faecorrigan:
I think for the next version I want to add another tab for info that includes all of the parameters and what they do and descriptions of some of the symbols. It would clean up some of the noise with the tooltips on bores/bosses as well

CHANGELOG.md
```diff
@@ -10,6 +11,7 @@
 - Bugfix: Fixed Speed/Feed scaling when using +/- buttons. Now can reach 10% and 300% scaling.
 - Bugfix: Fixed firmware download button. Now opens the github releases page for the Community firmware
 - Bugfix: Fixed squished center buttons
+-Enhancement: Added 3 axis probe screen
```
SergeBakharev, line 14:
```suggestion
- Enhancement: Added 3 axis probe screen
```

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 [Unreleased]
 - Enhancement: The ability to reduce the size of the autolevel probe area. This enables the ability to avoid probing the stock where there might be obstacles preventing probing.
 - Enhancement: tooltip support. See https://github.com/Carvera-Community/Carvera_Controller/pull/143 for more information
+-Enhancement: tooltip support added for switches and text input fields
```
SergeBakharev, line 4:
```suggestion
- Enhancement: tooltip support added for switches and text input fields
```

.gitignore
```diff
@@ -84,3 +84,9 @@ xcuserdata
 project.pbxproj*backup
 packaging_assets/ios/dist
 packaging_assets/ios/carveracontroller-ios/carveracontrollerpkg/*
+Angle-probe-settings.json
+bore-probe-settings.json
+Boss-probe-settings.json
+inside-corner-settings.json
+outside-corner-settings.json
+single-axis-probe-settings.json
```
SergeBakharev, line 92: What creates these files?

scripts/update_translations.py
```diff
@@ -13,7 +13,7 @@
 
 # Source files with UI text. This is analysed to create the .pot file
 PY_FILES = ["main.py", "Controller.py", "GcodeViewer.py"]  # Python source files
-KV_FILES = ["makera.kv"]  # .kv files
+KV_FILES = ["makera.kv","addons/probing/ProbingPopup.kv","addons/probing/preview/ProbingPreviewPopup.kv"]  # .kv files
```
SergeBakharev, line 16: Aren't there much more .kv files now?

.gitignore
faecorrigan: they persist settings between controller restart. and are created in the {probetype}settings.py files
it is something Todd setup

scripts/update_translations.py
faecorrigan: quite a few, should probably add the settings one into this section as well

carveracontroller/addons/probing/preview/ProbingPreviewPopup.py
```diff
@@ -0,0 +1,31 @@
...
+        super(ProbingPreviewPopup, self).__init__(**kwargs)
+
+    def get_probe_switch_type(self):
+        return 1
+        # if self.cb_probe_normally_closed.active:
+        #     return ProbingConstants.switch_type_nc
+        #
+        # if self.cb_probe_normally_open.active:
+        #     return ProbingConstants.switch_type_no
+        return 0
```
SergeBakharev, line 24: This is unreachable code.

faecorrigan: nope
missed that in my review, I can purge that section of code

carveracontroller/addons/probing/operations/SingleAxis/SingleAxisProbeOperationZAxis.py
```diff
@@ -0,0 +1,39 @@
+import copy
+
+from docutils.nodes import definition
```
SergeBakharev, line 3: docutils is not in the pyproject.toml as a dependency for this project

carveracontroller/addons/probing/preview/ProbingPreviewPopup.py
faecorrigan: I will need to check with todd about that whole preview section. I don't think it is being used but I want to make sure it isn't important to something he is working on

carveracontroller/addons/probing/operations/SingleAxis/SingleAxisProbeOperationZAxis.py
faecorrigan: what is the best way to resolve this?
is it ok to include docutils in the toml?

.gitignore
SergeBakharev: Got it, thanks. I'll make a note to have a closer look at how these files are being referenced/created. The PyInstaller packaged artifacts unpack into a randomly named temp directory as part of execution, so any settings files would need to be written to a predefined location like ~/.kivy/config

scripts/update_translations.py
SergeBakharev: Maybe instead of using a hard  ided list of files just do some globs? *.py, *.kV

I think I've removed all the unused files from the stock codebase

carveracontroller/addons/probing/operations/SingleAxis/SingleAxisProbeOperationZAxis.py
SergeBakharev: Easiest way is to use `poetry add docutils` as it will both add the definition to the pyproject.toml and add the hash values into the lock file.

But you also can just add the definition to the toml manually and run `poetry lock --regenerate` to get poetry to update the lock file.

The lock file ensures that we are using the same 3rd party dependency versions (going as far as checking the SHA hash incase the vender re-releases a modified version of a package with the same version strong).

.gitignore
SergeBakharev: The probe settings are being saved into the current working directory. That's pretty ugly, and inconsistent with the rest of the program.

We really should create a dedicated application config directory, and write some migration code to move off using ~/.kivy/config but I think that's a bigger piece of work for another time.

For now the probe settings should instead be written to the kivy config like all the other config variables.

toddmitchell: > The probe settings are being saved into the current working directory.  That's pretty ugly, and inconsistent with the rest of the program.

Totally agree that it should not be their final location, however, I intentionally did not save those in the main configuration for two reasons

1. this is an add-on so I don't want to conflict or have them overwritten by changes incoming from makera
2. I did not want to pass in the config instance from the main controller code since this is an add-on.

SergeBakharev: What exactly makes this an add-on? The codebase doesn't have any "plug-in" framework, and I don't see any feature-flagging code to disable or abstract this functionality. To me this code is no different to any other code in the project. 

From that perspective we have already added plenty of additional config keys to the .kivy/config file. You are right this has the potential for conflict or be overwritten if someone interchangeably uses the Makera and Community versions of the controller. But this hasn't occurred in a negative way yet. Makera frankly haven't added much functionality at all to their controller in the last 2 years, that's one of the reasons why this project exists. So I wouldn't be too worried about conflicts. 

We should mid-term plan to migrate to a unique config file/location of our own to further mitigate the risk. The .kivy/config file is actually the default for _any_ kivy application.

 As a compromise why don't you save the files in the "~/.kivy/" directory. At least that way the files are out of sight, and in the same directory as the other config file.

carveracontroller/addons/probing/ProbingPopup.kv
```diff
@@ -0,0 +1,505 @@
...
+            height: '40dp'
+            halign: 'center'
+            Label:
+                size_hint_min_x: dp(90)
+                padding: '5dp'
+                halign: 'center'
+                valign: 'middle'
+                text_size: self.size
+                multiline: False
+                text: tr._('Note: Most of these operations require a true 3 axis probe, not the block supplied by Makera. See the community youtube page for more info')
```
SergeBakharev, line 496: Makera refer to the probing block in their manuals as the "[manual probe](https://wiki.makera.com/en/carvera/manual/tool-kit#manual-probe)", we should stick with their nomenclature since that's what the user is most likely to be familiar.

```suggestion
                text: tr._('Note: Most of these operations require a true 3 axis probe, not the manual probe supplied with the Carvera. See the community youtube page for more info')
```

carveracontroller/addons/probing/operations/SingleAxis/SingleAxisProbeOperationZAxis.py
faecorrigan: docutils is not actually used and I ran into a lot of issues adding it to the project with version mismatches
So I removed the imports

scripts/update_translations.py
faecorrigan: I am going to do this as a separate PR and just remove the stuff from the probing screen for now. I should have that PR done by the end of the weekend

.gitignore
faecorrigan: I pushed a change to the directory to save in the user folder. @SergeBakharev can you double check my work?

carveracontroller/addons/probing/ProbingPopup.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def delayed_bind(self, dt):
+        self.outside_corner_settings = self.ids.outside_corner_settings
+        self.inside_corner_settings = self.ids.inside_corner_settings
+        self.single_axis_settings = self.ids.single_axis_settings
+        self.bore_settings = self.ids.bore_settings
+        self.boss_settings = self.ids.boss_settings
+        self.angle_settings = self.ids.angle_settings
+
+    # def on_bore_boss_corner_probing_pressed(self, operation_key: str):
```
SergeBakharev, line 53:
```suggestion
```

carveracontroller/addons/probing/ProbingPopup.py
```diff
@@ -0,0 +1,105 @@
...
+            self.preview_popup.gcode = gcode
+            self.preview_popup.probe_preview_label = gcode
+        else:
+            self.preview_popup.gcode = ""
+            self.preview_popup.probe_preview_label = "Missing required parameter " + missing_definition.label
+
+        self.preview_popup.open()
+
+    def load_config(self):
+        # todo
```
SergeBakharev, line 104: Was there meant to be something here?

carveracontroller/addons/probing/ProbingPopup.py
```diff
@@ -0,0 +1,105 @@
...
+        pass
```
SergeBakharev, line 105: Was there meant to be something here?
```suggestion
```

SergeBakharev:
I get these warnings when I run this branch:
```
[WARNING] [Factory     ] Ignored class "ProbeSettingLabel" re-declaration. Current -  module: None, cls: <class 'carveracontroller.addons.probing.ProbingControls.ProbeSettingLabel'>, baseclass: None, filename: None. Ignored -  module: None, cls: None, baseclass: GridLayout, filename: REDACTED\carveracontroller\addons\probing\ProbingControls.kv.
[WARNING] [Lang        ] addons/tooltips/Tooltips.kv has already been included!
[WARNING] Deprecated property "<BooleanProperty name=allow_stretch>" of object "<kivy.uix.image.Image object at 0x0000021088E965F0>" was accessed, it will be removed in a future version
[WARNING] Deprecated property "<BooleanProperty name=keep_ratio>" of object "<kivy.uix.image.Image object at 0x0000021088E965F0>" was accessed, it will be removed in a future version
```

.gitignore
SergeBakharev: Looks good to me

toddmitchell: > What exactly makes this an add-on? The codebase doesn't have any "plug-in" framework, and I don't see any feature-flagging code to disable or abstract this functionality. To me this code is no different to any other code in the project.

Sorry for my slow reply.. I didnt catch any notification on this.  this looks all done so thanks @faecorrigan for picking it up.  I've been on the road,etc, etc.

anyway, the thing that makes it an add-on is exactly the risk you posed.  you have to merge code from makera, even if they dont make many updates.  I have zero opinion on where these files really go as I'm a occasional contributor.  as such, im happy to following any standards you all put in place.

PR #225, opened 2025-06-09
Fixes for iOS version of the app

zittix:
- Fix upload of a file in a remote non-writeable dir (also fixes the OS X version that has similar issue)
- Fix MDI window showing the keyboard for onscreen keyboard devices


SergeBakharev:
Thanks for this, putting the compressed .lz files into the temp directory is a good move. I would be tempted to do it always, but lets see how this goes for now.

PR #232, opened 2025-06-11
Add initial support for WHB04 pendant

yaqwsx:
Hi,

I am really used to MPGs for CNCs. I implemented initial support for the cheap Chinese WHB04 wireless MPG into the Controller. At the moment, I consider this a proof of concept rather than a polished solution. But I would like to gather feedback on the usage sooner rather than later.

## What works:

- basic driver of the pendant:
	- we can read all user inputs and
	- we can control the LCD to some extent.
- jogging of each axis with displaying of the workpiece coordinates on the LCD
- you can turn on/off the spindle.
- you can trigger e-stop via the RESET button.

## What doesn't work:

- I didn't manage to properly RE the display protocol of the pendant. So I struggle to properly set display flags,  e.g., the step size.
- Jogging shouldn't work when the machine is running or the jogging UI is not shown.
- Feed and spindle overrides are not implemented yet. 
- Job control is not implemented yet.
- It is undecided what buttons to support and what they should do.
- It works on Linux, I haven't tested Windows yet. It will probably crash on Android and iOS.

## How to test it:
- just plug

At the moment, I am trying to dive into the code and understand how the code is structured and how to properly hook into it. Any feedback in this regard is welcome! I also appreciate any suggestions on what the control elements should do.

Screenshot from 2025-06-11 21-49-04
Screenshot from 2025-06-11 21-48-51

https://github.com/user-attachments/assets/64410206-d979-47f9-8b8c-653cc2d9736f

### DIFF
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -296,9 +296,12 @@ class Controller:
     def clearAutoLeveling(self):
         self.executeCommand("M370\n")
 
-    def setSpindleSwitch(self, switch, rpm):
+    def setSpindleSwitch(self, switch, rpm=None):
         if switch:
-            self.executeCommand("M3 S%d\n" % (rpm))
+            if rpm is None:
+                self.executeCommand("M3\n")
+            else:
+                self.executeCommand("M3 S%d\n" % (rpm))
         else:
             self.executeCommand("M5\n")
 
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -200,11 +200,11 @@ class Controller:
                 cmd = "buffer " + cmd
             self.executeCommand(cmd) #run margin command. Has to be two seperate commands to offset the start of the autolevel process
         cmd = "M495 X%gY%g" % (CNC.vars['xmin'] + auto_level_offsets[0], CNC.vars['ymin'] + auto_level_offsets[2]) #reinitialize command with any autolevel offsets
-        if zprobe: 
-            if zprobe_abs: 
+        if zprobe:
+            if zprobe_abs:
                 cmd = "M495 X%gY%g" % (CNC.vars['xmin'], CNC.vars['ymin']) #reset command for 4th axis
                 cmd = cmd + "O0"
-            else: 
+            else:
                 cmd = cmd + "O%gF%g" % (z_probe_offset_x, z_probe_offset_y)
         if leveling:
             cmd = cmd + "A%gB%gI%dJ%dH%d" % (CNC.vars['xmax'] - (CNC.vars['xmin']+auto_level_offsets[1]+ auto_level_offsets[0]) , CNC.vars['ymax'] - (CNC.vars['ymin']+auto_level_offsets[3] + auto_level_offsets[2]), i, j, h)
@@ -384,7 +384,7 @@ class Controller:
 
     def clampToolCommand(self):
         self.executeCommand("M490.1\n")
-    
+
     def unclampToolCommand(self):
         self.executeCommand("M490.2\n")
 
@@ -829,8 +829,8 @@ class Controller:
 
     # ----------------------------------------------------------------------
     def jog(self, _dir):
-        self.executeCommand("G91G0{}".format(_dir))
-    
+        self.executeCommand("G91G0{}F5000".format(_dir))
+
     def jog_with_speed(self, _dir, speed):
         if speed > 0:
             self.executeCommand(f"G91G0{_dir} F{speed}")
@@ -845,6 +845,17 @@ class Controller:
         if y is not None: cmd += "Y%g" % (y)
         if z is not None: cmd += "Z%g" % (z)
         self.sendGCode("%s" % (cmd))
+
+    def gotoSafeZ(self):
+        self.sendGCode("G53 G0 Z0")
+
+    def gotoMachineHome(self):
+        self.gotoSafeZ()
+        self.sendGCode("G53 G0 X0 Y0")
+
+    def gotoWCSHome(self):
+        self.gotoSafeZ()
+        self.sendGCode("G53 G0 X%g Y%g" % (CNC.vars['wcox'], CNC.vars['wcoy']))
 
     def wcsSetA(self, a = None):
         cmd = "G92.4"
@@ -886,7 +897,7 @@ class Controller:
         cmd += pos
 
         self.sendGCode(cmd)
-    
+
     def wcsClearRotation(self):
         cmd = "G10L2R0P0"
         self.sendGCode(cmd)
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -297,13 +297,13 @@ class Controller:
         self.executeCommand("M370\n")
 
     def setSpindleSwitch(self, switch, rpm=None):
-        if switch:
-            if rpm is None:
-                self.executeCommand("M3\n")
-            else:
-                self.executeCommand("M3 S%d\n" % (rpm))
+        if switch and rpm is not None:
+            cmd = f"M3 S{int(rpm)}\n"
+        elif switch and rpm is None:
+            cmd = "M3\n"
         else:
-            self.executeCommand("M5\n")
+           cmd = "M5\n"
+        self.executeCommand(cmd)
 
     def setVacuumPower(self, power=0):
         if power > 0:
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -829,7 +829,7 @@ class Controller:
 
     # ----------------------------------------------------------------------
     def jog(self, _dir):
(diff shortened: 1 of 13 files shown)


SergeBakharev:
FYI I've changed the PR destination to our `develop` branch, this is where we stage changes that are unreleased.

Vignesh1230:
image

Windows build doesnt work off the bat. Will debug as well

Vignesh1230:
image
Only when running in the poetry environment in dev

SergeBakharev:
> image
> 
> Windows build doesnt work off the bat. Will debug as well

Per the [Python hid package README](https://pypi.org/project/hid/) the library is a wrapper for [hidapi](https://github.com/libusb/hidapi) which itself supports a number of different possible backends. The docs also say that **hidapi** needs to be installed separately. On Windows it uses libusb/hidapi (hid.dll) which has binaries [here](https://github.com/libusb/hidapi/releases).

@yaqwsx Is there a reason why you didn't use the pre-existing hid input support from [kivy](https://kivy.org/doc/stable/_modules/kivy/input/providers/hidinput.html)? It's certainly easy enough to provide the dependencies for the library you have used to PyInstaller as part our build automation to have it include the binary artifact, but it's duplicate functionality to the existing UI framework and thus extra complexity and dependency management.

yaqwsx:
@SergeBakharev: The reason why I haven't used kivi is my lack of knowledge. This is the first time I've seen kivy, so I wasn't aware it has support for HID devices. I will have a look at it and I think I can make it work with it. The only reason why I used hidapi is that I have a have a previous experience with it.

In the meantime, I have rebased my branch on top of develop, I am working on implementing jogging limits, and I have a few ideas to improve the responsiveness.

carveracontroller/makera.kv
```diff
@@ -3502,7 +3502,11 @@
                                                 on_release: app.root.controller.setLightSwitch(not self.state=="normal")
 
                                         GridBoxLayout:
-                                            # Empty for now
+                                            ToggleButton:
```
SergeBakharev, line 3505: I suggest moving this button to the "advanced_jog_controls" button group since this spot in the existing grid has been taken up by the `Ext. Control` toggle button (line3515 in makera.kv)

carveracontroller/addons/pendant/pendant.py
```diff
@@ -0,0 +1,102 @@
...
+from ...Controller import Controller
+from . import whb04
+
+from kivy.clock import Clock
+from kivy.uix.settings import SettingItem
+from kivy.uix.spinner import Spinner
+from kivy.uix.togglebutton import ToggleButton
+
+
+class Pendant:
```
SergeBakharev, line 12: Love the future thinking to support multiple different Pendant types!

carveracontroller/makera.kv
SergeBakharev: I'm not so keen on having it so prominently displayed when it requires additional hardware most users won't have, and when there is already a jog controls UI section. If you can determine how to only show the button conditionally if the feature flag is enabled in the Controller settings that would be ideal.

Excuse my ignorance but why is a button to activate pendent jogging required at all? Doesn't the pendent itself have a power button, and wouldn't it be more convenient to use that?

yaqwsx:
You plug the pendant into a computer. There is no need for custom firmware, there is a need for the controller version from this pull request.

yaqwsx:
I validated that the solution works on Windows. The requirement is that `hidapi.dll` is located on `PATH` or in the directory with  `python.exe` or in the `System32` directory.

I adjusted the code such that failure in loading a pendant driver doesn't stop the controller from running.

CCS86:
> You plug the pendant into a computer. There is no need for custom firmware, there is a need for the controller version from this pull request.


I'm not super experienced with Github. How do I download the compiled version of the controller with your modifications, since there are no releases on your fork? Do I have to download the files from https://github.com/yaqwsx/Carvera_Controller/tree/pendant and compile myself?

yaqwsx:
I have nearly finished the pendant integration. The only thing it misses is the ability to detect if the machine is running a file or not. Based on this, the pendant should be enabled/disabled, and it should also drive the behavior of the start/pause button. 

These are small changes, though. Overall, I think the code is ready to test and for a code review.

carveracontroller/__main__.py
```diff
@@ -1,9 +1,5 @@
-from carveracontroller.main import main, init_lang, Lang
-
-# tr is used throughout the kivvy .kv definition files
-# thus kivy expects it be availiable from the caller
-default_lang = init_lang()
-tr = Lang(default_lang)
+from carveracontroller.main import main
+from carveracontroller.translation import tr
```
mcowger, line 2: I'm not sure this change has the same effect?  Why was this changed?

carveracontroller/addons/pendant/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .pendant import SettingPendantSelector, SUPPORTED_PENDANTS, \
```
mcowger, line 1: Would recommend avoiding relative imports

carveracontroller/addons/pendant/pendant.py
```diff
@@ -0,0 +1,216 @@
+import json
+from typing import Callable
+
+import logging
+logger = logging.getLogger(__name__)
+
+from ...CNC import CNC
+from ...Controller import Controller
```
mcowger, line 8: Would recommend avoiding relative imports

carveracontroller/addons/pendant/pendant.py
```diff
@@ -0,0 +1,216 @@
...
+            logger.error(f"Failed to run macro {macro_id}: {e}")
+
+
+class NonePendant(Pendant):
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+
+try:
+    from . import whb04
```
mcowger, line 88: relative imports.

carveracontroller/makera.kv
```diff
@@ -3510,7 +3510,10 @@
                                                 on_release: app.root.controller.setExternalControl(100 if self.state=="down" else 0)
 
                                         GridBoxLayout:
-                                            # Empty for now
+                                            ToggleButton:
+                                                id: pendant_jogging_en_btn
+                                                text: tr._('No Pendant Connected')
```
SergeBakharev, line 3515: This text is quite long and goes beyond the button without a very large window. Maybe drop `Connected`:

```suggestion
                                                text: tr._('No Pendant')
```

carveracontroller/Controller.py
```diff
@@ -843,6 +846,17 @@ def goto(self, x=None, y=None, z=None):
...
+        self.sendGCode("G53 G0 Z0")
+
+    def gotoMachineHome(self):
+        self.gotoSafeZ()
+        self.sendGCode("G53 G0 X0 Y0")
+
+    def gotoWCSHome(self):
+        self.gotoSafeZ()
+        self.sendGCode("G53 G0 X%g Y%g" % (CNC.vars['wcox'], CNC.vars['wcoy']))
+
```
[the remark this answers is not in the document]
yaqwsx: I will respect your wish, but I personally consider gotoPosition, i.e., a function that takes a string and then, based on it, does different things, as a huge-huge antipattern.

carveracontroller/addons/pendant/__init__.py
yaqwsx: Can you provide rationale? I consider the relative import here pretty idiomatic idiomatic. I don't care where the absolute package is located, but I want to extract names from files into the package scope. It expresses the the idea well. Meanwhile `from carveracontroller.addons.pendant.pendant import ....` looks weird.

carveracontroller/addons/pendant/pendant.py
yaqwsx: You are right, that here importing from `carveracontroller` is more idiomatic. Will change.

carveracontroller/Controller.py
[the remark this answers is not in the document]
SergeBakharev: Actually you know on balance, I agree with you. It's a pain to remember all the possible gotoPosition options, its fragile, and can get pretty yucky quickly.

Leave what you have, and I'll replace gotoPosition() with the style you are using with separate methods in a separate PR.

carveracontroller/Controller.py
```diff
@@ -826,8 +829,8 @@ def grblRestoreAll(self):
 
     # ----------------------------------------------------------------------
     def jog(self, _dir):
-        self.executeCommand("G91G0{}".format(_dir))
-    
+        self.executeCommand("G91G0{}F5000".format(_dir))
```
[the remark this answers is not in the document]
yaqwsx: After studing the code a bit I decided to use `jog_with_speed` with a hard-coded value (with a comment giving rationale).

yaqwsx:
I reflected your comments and suggestions.

PR #236, opened 2025-06-18
Add drop down to 3D probing to select WCS zeroing of X/Y or X/Y/Z

SergeBakharev:
Add drop down to 3D probing to select WCS zeroing of X/Y or X/Y/Z not just X/Y to Inside/Outside Corner and Boss probing screens.

Also:
Fix: Invisible jog control panel buttons being clickable when panel disabled.

image

### DIFF
--- a/carveracontroller/addons/probing/ProbingControls.py
+++ b/carveracontroller/addons/probing/ProbingControls.py
@@ -1,6 +1,6 @@
 from kivy.uix.gridlayout import GridLayout
 from kivy.properties import StringProperty
-from ...addons.tooltips.Tooltips import ToolTipButton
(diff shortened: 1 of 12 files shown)


carveracontroller/addons/probing/ProbingControls.py
```diff
@@ -11,4 +11,7 @@ def __init__(self, image='', halign='center', valign='middle', **kwargs):
     
     def on_size(self, *args):
         pass
-        #at some point add code to respect halign and valign in img
\ No newline at end of file
+        #at some point add code to respect halign and valign in img
+
+class ProbeZeroXYZDropDown(ToolTipDropDown):
```
SergeBakharev, line 16: Neither is this

```suggestion
```

Issue #193, opened 2025-05-13
Restart from line X in Configure and Run UI


SergeBakharev:
Suggestion from FB user "Rob Ligeti":
> It would be awesome if the controller SW had a "Restart from line X" function or a save state for big jobs..
> Please, and thank you.

I think it would be a good feature to add to the Community Controller. Start running the gcode up until Spindle On (`M3` command) then pause and `goto` a specified line. The UI could even suggest lines with tool or height changes.

acj:
Will be implemented by #246

PR #244, opened 2025-06-21
Start Gcode playback at line in UI

SergeBakharev:
Implements #193


SergeBakharev:
Basic scenario with human entered line number works, but the alignment of the checkbox is off. Nothing obviously wrong in the kv file.... Will have more of play later

image

SergeBakharev:
Better 
image

PR #245, opened 2025-06-21
Revert "Start Gcode playback at line in UI"

SergeBakharev:
Reverts Carvera-Community/Carvera_Controller#244


SergeBakharev:
Start at line feature doesn't start the spindle which makes this dangerous to use as-is. Reverting

PR #246, opened 2025-06-21
Start at line functionality in UI

SergeBakharev:


SergeBakharev:
What is currently present works but since the Pause/goto are so clean the spindle is never started.

TODO:

- [ ] Parse Gcode checking for any incremental movement modes. If found abort, this feature won't work
- [ ] Parse Gcode checking for metric/inch mode and insert start before <resume>
- [ ] Parse Gcode finding closest `M3` and `M6` commands to start line, and insert that before <resume>

SergeBakharev:
This is ready for review. It adds 4 enhancements that are inter-related:

1. Gcode file simulation progress bar is now synchronised with the line selection in the gcode file. Selecting a line jumps the visualisation to that point, and vice versa.

2. Config and Run screen has an extra option to resume playback to line number. Progressing with this option enabled open a prompts with a preview of the resume commands.

3. GcodeViewer now has a context menu when you right click or long press a line number. Only options are select line for resume and cancel.

4. Halting or stopping a gcode playback automatically populates the resume line number to the last played line, but does not select the resume check box.

Since I wrote half of this 6 months back, and with AI assistance I'm going to go through it myself in a few days as well. It also needs to be tested with more scenarios, I've mostly been poking it with the Example files. It supports laser and spindle modes, and all 4 axis.

carveracontroller/main.py
```diff
@@ -5049,6 +5284,40 @@ def enable_laser_mode_confirm_popup(self):
         self.confirm_popup.cancel = None
         self.confirm_popup.open(self)
 
+    # -----------------------------------------------------------------------
+    def open_resume_playback_confirm_popup(self, file_name, start_line):
+        if self.confirm_popup.showing:
+            return
+        
+        # Check if the file contains G91 (incremental movement) commands
```
SergeBakharev: that's a old comment removing
```suggestion
```

carveracontroller/Controller.py
```diff
@@ -549,6 +560,522 @@ def playCommand(self, filename):
...
+        while l <= r:
+            mid = (l + r) >> 1
+            if array[mid] >= key:
+                ans = mid
+                r = mid - 1
+            else:
+                l = mid + 1
+        return ans - 1
+
+    def _get_line_position_from_gcode_viewer(self, line_number):
```
SergeBakharev: This is a bit hacky, I didn't want to have more than one gcode parser, so I hook into the existing gcode_viewer. Need to have a think if this can be done more cleanly.

carveracontroller/Controller.py
```diff
@@ -549,6 +560,522 @@ def playCommand(self, filename):
...
+                # Look for M3 that's not followed by a digit (to avoid M30, M31, etc.)
+                # M3 can be preceded by anything (including digits from other commands like S12000). 
+                # Yes, really! One of the example files (ACRYLIC-R2D2.nc) has "G0X0.000Y0.000S12000M3" wtf is that!!?
+                m3_match = re.search(r'M3(?![0-9])', line_upper)
+                if m3_match:
+                    most_recent_m3_line = i
+                    # Extract S parameter from this line (S can appear before or after M3)
+                    # Use a more robust pattern that handles S parameters anywhere in the line
+                    # Pattern: S followed by optional sign, digits, optional decimal point and more digits
+                    s_match = re.search(r'[Ss]([+-]?\d+(?:\.\d+)?)', line)
```
SergeBakharev: Probably a bit too open. Should just be S with an integer. I hate overly complicated regex.

carveracontroller/main.py
```diff
@@ -275,13 +275,20 @@ def send_mdi_command(self):
         app.root.send_cmd()
 
 class GcodePlaySlider(Slider):
+    def __init__(self, **kwargs):
+        super(GcodePlaySlider, self).__init__(**kwargs)
+        self.programmatic_update = False  # Flag to prevent infinite loops
```
SergeBakharev: Need to review if this is still required. Originally it was to track if a slider was last updated by the App itself (because of line selection) or because of a user input (and thus the line selection needed to be updated).

acj:
So glad to see this feature being added. :tada: Here are my notes from testing on the CA1, in no particular order:

- Sometimes the resume line is slightly later than what gcode viewer is showing, e.g. resuming on L123 while the viewer has L121 selected. It's usually been 1-2 lines off for me.
- After an e-stop, I resumed my program. It started the spindle and then tried to do a tool change that it had already done a moment before. I clicked confirm since the tool was already loaded, and then it started to measure TLO with the spindle running. I haven't been able to repro but will keep trying. Maybe there's a more general safety fix we can do in the TLO code?
- When resuming, sometimes it starts the spindle and then retracts, and other times it retracts and then starts the spindle. Seems odd that it's inconsistent. Retracting first seems safer. It also seems like it retracts much farther than it needs to.
- If you use the resume feature and then let a program finish, opening the Config and Run screen will try to resume near the last line of the prog. I'd suggest clearing the checkbox/number on program completion
- The gcode viewer gets out of sync and sometimes jumps forward and backward in a way that's hard to understand. This might be a general problem with that UI component rather than something specific to resuming, though. See the attached video starting at about 0:28 (but also note that the wrong line is highlighted when it prompts for the tool change). I was running a two-step program generated by MakeraCAM that mills out the head of a bolt and uses two different tools. It seemed to be in sync for the first half and then got much worse after the tool change. Happy to share the program if it's helpful.
- After using the context menu to select a line to resume on, it would be nice to show a brief toast so that you know it worked

I should have time to look over the code tomorrow-ish.


https://github.com/user-attachments/assets/80dc3c67-8986-4288-b26c-32c24b1c6dae

SergeBakharev:
There seems to be an "off by one" error either in the controller, the firmware, or my understanding of `goto`.

In the gcode example below, resuming to line 9 does not restart the spindle. Seems like `goto` is going to the end of line 9 (actually line 10).

image

The resume injected gcode is pretty sparse since at line 9 there is not yet a spindle speed or movement position.
image

SergeBakharev:
Thanks for the detailed feedback and video @acj I'll investigate further too. Some of this inconsistent behaviour is definitely coming from the firmware. I experienced a similar issue with the execution of "goto" and the buffered M600. My original implementation didn't have the Controller waiting for the paused state before running the other commands but I found it to work inconsistently. I think there is some kind of race condition occurring in the firmware. Not sure if we should find and resolve the issue in the firmware (might be a can of worms) or do more defensive command execution in the controller similar to how I have it wait for the machine status change to paused after running `play` before executing the next few commands. Might be enough to just wait a literal second or two between command executions and call it a day.

The jumping around line selection probably is related to the change I made to keep the line and progress bar selection in sync, that functionality should only be used when in "simulation" mode not when running a file. Ultimately I want to throw away the "synchronised" visualisation stuff (keeping it only for "preview" purposes) and just move to showing the actual spindle motion/location based on machine status. It will be smoother, and more accurate overall. But I don't want to involve any more changes in this PR.

carveracontroller/makera.kv
```diff
@@ -5626,4 +5623,39 @@
...
+    BoxLayout:
+        orientation: 'vertical'
+        padding: 0
+        spacing: 0
+        pos_hint:{'top': 1,'right':1}
+        
+        # Resume playback option
+        Button:
+            id: resume_option
+            text: 'Select this line for resume playback'
```
acj: Maybe clearer
```suggestion
            text: 'Resume at this line'
```

carveracontroller/main.py
```diff
@@ -5049,6 +5284,46 @@ def enable_laser_mode_confirm_popup(self):
...
+    # -----------------------------------------------------------------------
+    def open_resume_playback_confirm_popup(self, file_name, start_line):
+        if self.confirm_popup.showing:
+            return
+        
+        app = App.get_running_app()
+        local_file_path = app.selected_local_filename if hasattr(app, 'selected_local_filename') else None
+        
+        # Get command preview from Controller
+        commands = self.controller.playStartLineCommand(file_name, start_line, preview=True, local_file_path=local_file_path)
```
acj: Does the file path have ascii squares in it for you? I'll get a screenshot next time I'm at the machine. There might be a simple string replace we can do to make it look more polished.

carveracontroller/Controller.py
```diff
@@ -549,6 +560,522 @@ def playCommand(self, filename):
...
+                
+                # Check if line contains G1/G01/G2/G02/G3/G03
+                line_upper = line.upper()
+                # Check for G1/G01/G2/G02/G3/G03 commands
+                # These can appear at start of line or anywhere in the line
+                g_commands = ['G1', 'G01', 'G2', 'G02', 'G3', 'G03']
+                found_g_command = False
+                for cmd in g_commands:
+                    # Check if command appears (at start or after non-alphanumeric char, not followed by digit)
+                    if line_upper.startswith(cmd):
```
acj: This doesn't match the comment above, specifically the "not followed by digit" part. Do we want a similar check to L837 below?

acj:
Still seeing the gcode viewer jump around and get out of sync, unfortunately. I'm running the latest dev firmware (https://github.com/Carvera-Community/Carvera_Community_Firmware/commit/a79231f89054dda2cafae5bcfb76775b567d8edc).

Things I noticed:
- Love the green flag 👍
- When I have my two-tool test program loaded and try to resume halfway through on the "T5 M6" command (T5 is already loaded), instead it loads the other tool (T4), moves to work origin, and then goes back to load T5. But if I tell it to resume two lines AFTER "T5 M6", then it correctly chooses T5.
- Gcode viewer's selected line is ~2 lines behind what the machine is actually doing
- The resume line number on the Config and Run dialog isn't reset when you load a new program, so it can be a number that doesn't make sense anymore
- Bottom progress bar is usually 3-5 lines ahead of the gcode viewer's selected line (might be unrelated to your changes)
- Still seeing backwards jumps in gcode viewer. Sometimes the bottom progress bar jumps backward too
- When an X-line program finishes, it puts the green flag at line X-2. (Still feels odd to me that it uses the end of the program as a "resume from" value)

I'm short on time tonight and will play around more this weekend.

SergeBakharev:
Thanks @acj for your detailed notes.

> When I have my two-tool test program loaded and try to resume halfway through on the "T5 M6" command (T5 is already loaded), instead it loads the other tool (T4), moves to work origin, and then goes back to load T5. But if I tell it to resume two lines AFTER "T5 M6", then it correctly chooses T5.

My bad, I put in logic to adjust the start_line number but had it apply to the entire code generation method instead of just the goto command generation.

> Gcode viewer's selected line is ~2 lines behind what the machine is actually doing

It's unavoidable I think, the Controller only polls the machine state every 200ms.

> The resume line number on the Config and Run dialog isn't reset when you load a new program, so it can be a number that doesn't make sense anymore.

How did I miss that, there is code that clear it, but it wasn't working. Fixed now.

> Bottom progress bar is usually 3-5 lines ahead of the gcode viewer's selected line (might be unrelated to your changes)

Good spot. It is indeed unrelated to my changes, but we can try to fix it up here. I found that two different sources of truth are used which update at different times. I've put int some code to sync cur_line_index with display_count. Will test this further over the next few days

> Still seeing backwards jumps in gcode viewer. Sometimes the bottom progress bar jumps backward too

I'll investigate this further with the newly added [comms logging feature](https://github.com/Carvera-Community/Carvera_Controller/pull/480) I'm suspicious that maybe the machine itself is returning status messages that are either out of order or due to some bug containing previous line numbers.

> When an X-line program finishes, it puts the green flag at line X-2. (Still feels odd to me that it uses the end of the program as a "resume from" value)

Ah, that's a fun one. The machine can complete gcode playback before the Controller polls the status of it running the last line. So the last update the Controller gets is X-2, then the machine state changes to Idle. I don't think there is any way to tell that a gcode file has successfully completed executing (except the line `Done printing file` echoed on the serial interface). The best solution is for the firmware to be enhanced so that the machine status response retains the playback details after completion (showing 100% complete etc) until it starts playing the next file. But even if we were to add that we would need to have an alternative method of determining playback success to retain compatibility with the Makera firmware. Maybe we just assume that if we reach 98% completion or line number max()-5 that the file was successfully executed?

acj:
> Maybe we just assume that if we reach 98% completion or line number max()-5 that the file was successfully executed?

Those both seem reasonable to me, especially the latter. It might get weird if you're working with a very short program, but in that situation it's easy to right-click and choose the line you want.

I should have time tomorrow to try out your latest changes.

acj:
Testing on 566d23f:
- The gcode viewer isn't auto-scrolling at all when the program is running. The resumed-at line stays selected and in the same position on the screen
- The bottom toolbar and progress bar aren't working when the program is running. The elapsed time counter ticks away, but there's no progress % and the pause/stop/etc buttons are disabled. I even restarted the machine a couple of times because this issue and the one above seemed so odd. It's consistent, though.
- Letting the program finish doesn't clear the resume-at-line value. Maybe related to the issues above?
- If I put the flag on "T5 M6" and run the program, it still tries to load T4. Resuming at the very next line will load T5, though. 50% closer 😄

SergeBakharev:
I couldn't help myself with the [firmware addition ](https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/227) and had been running that without the same progress bar/resume issues. Clearly I ran before I walked. Dang it, sorry about that @acj , I'll try to get it work correctly without the firmware status change.

SergeBakharev:
Okay, the other issues were because I jumped the gun and put in logic that relied on https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/227 being merged. I've commented it out for now, and the progress/playback is now fixed

acj:
Sounds good. I'll give it another spin tonight or tomorrow

acj:
The latest changes look really good on CA1. I think I re-tested all the things we've talked about.

I got curious about the forward/backward line jumping and added some logging. As you mentioned earlier, it's actually coming from the machine. Not the controller's fault. I _think_ it's a subtle bug in `Player::on_get_public_data` but need to do more testing. I'll make a PR if I can figure out a reliable fix.

Anyway, this seems good enough to merge so that we can get more folks using it. Nice job!

PR #264, opened 2025-06-26
Fixed the error after reconnecting the controller (Load config error)

WARIO2412:
Hope this fix is right...


SergeBakharev:
This looks right

PR #266, opened 2025-06-27
Clear rotation for 4th axis and bug fixing after last changes

WARIO2412:


carveracontroller/Controller.py
```diff
@@ -865,6 +883,10 @@ def wcsSetM(self, x = None, y = None, z = None, a = None):
         cmd += pos
 
         self.sendGCode(cmd)
+    
+    def wcsClearRotation(self):
+        cmd = "G10L2R0P0"
```
SergeBakharev, line 888: What happens when this is run on stock firmware?

carveracontroller/main.py
```diff
@@ -553,6 +554,13 @@ def __init__(self, config, **kwargs):
 
         # Ensure the spinner is updated after initialization
         Clock.schedule_once(self.populate_spinner, 0)
+    
+    def on_open(self):
+        super().on_open()
+        app = App.get_running_app()
+        if app.has_4axis:
+            self.controller.wcsClearRotation()
```
SergeBakharev, line 562: This essentially sends a GCode command to clear the rotation when the config-and-run screen is opened.

Is there a reason to do it specifically at this time?

I think it would be safer to do this once the user is intending to actually run the gcode file.

Specifically I think in the `Makera.apply()` method is best, as this run before `Makera.playCommand()` (see `Makera.play()`) and also run if doing only a zprobe/margin etc.

SergeBakharev (changes requested):
The parseBracketAngle() changes look fine but I think the other two changes could be put in better places.

carveracontroller/Controller.py
WARIO2412: Nothing, it's just ignored

carveracontroller/main.py
```diff
@@ -3113,6 +3123,14 @@ def updateStatus(self, *args):
...
             self.heartbeat_time = now
             app = App.get_running_app()
+            if app.model!= CNC.vars['MachineModel'] and CNC.vars['MachineModel']!= 999:
+                machinemodel = ''
+                if CNC.vars['MachineModel'] == 1:
+                    machinemodel = 'C1'
+                elif CNC.vars['MachineModel'] == 2:
+                    machinemodel = 'CA1'
+                if app.model!= machinemodel:
+                    Clock.schedule_once(partial(self.setUIForModel, machinemodel), 0)
```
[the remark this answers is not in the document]
WARIO2412: Done. The stock regex code snippet should work

carveracontroller/main.py
WARIO2412: Changed it to the apply() method and changed it that way, that the rotation is not applied to the coord popup when the 4th is active

PR #277, opened 2025-07-03
added armv8 and x86_64 support to the apk package

SergeBakharev:
This should cover 99% of the android devices on the market, based on that I've dropped the arch from the filename


WARIO2412 (approved):
Looks good to me

PR #278, opened 2025-07-03
fixed the set origin popup to calculate the offsets on pressing the different options

WARIO2412:


carveracontroller/main.py
```diff
@@ -331,6 +334,27 @@ def selected_anchor(self):
         elif self.cbx_current_position.active:
             return 4
         return 1
+    
+    def update_offsets(self):
+        # Use the same logic as CoordPopup.load_origin_label to set offsets
+        app = App.get_running_app()
```
SergeBakharev, line 340: As you pointed out you can unselect a radio button (I couldn't find a kivy attribution to prevent that).

If you unselect the radio button option it still triggers `update_offsets()` but since there is no active selection the value of `x` and `y` are unset and causes an app crash.

Lets just set an initial value for x and y to prevent this scenario.

```suggestion
        app = App.get_running_app()
        x = 0
        y = 0
```

SergeBakharev (approved):
If just the one issue with the radio unselect crash. Once that is handled, please add a line in CHANGELOG.md and this PR is good to merge

PR #282, opened 2025-07-14
Pendant Button behaviour and mcs coordinate changes

WARIO2412:
This PR fixes the machine coordinates for the machine home functions (going to XYZ -1 instead of 0)

The machine can now also be unlocked with the pause resume key, if in an alarm state.
The Probe Z button of the Pendant actually probes Z (up for discussion). With the community firmware it is done with M466 Z-200 S2 (setting the Z origin directly). With the stock firmware it's a G38.2 Z-200.

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -1908,8 +1908,8 @@ class Makera(RelativeLayout):
 
         # Setup pendant
         self.setup_pendant()
-        self.pendant_jogging_default = int(Config.get('carvera', 'pendant_jogging_default'))
-        self.pendant_probe_z_alt_cmd = int(Config.get('carvera', 'pendant_probe_z_alt_cmd'))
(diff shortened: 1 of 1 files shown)


yaqwsx:
Hi! Sorry for my limited reply - I am away from my office for ~2 weeks. Overall, I think the changes are OK.

Regarding probing, I have no strong opinion. I find the probing screen to be more generic. However, such a screen still requires interaction with the controller. What about we add a setting to configure the behavior?

WARIO2412:
> Hi! Sorry for my limited reply - I am away from my office for ~2 weeks. Overall, I think the changes are OK.
> 
> Regarding probing, I have no strong opinion. I find the probing screen to be more generic. However, such a screen still requires interaction with the controller. What about we add a setting to configure the behavior?

I understand why you chose the probing screen. Setting sounds good to me! I'll change it

SergeBakharev (approved):
Looks and tests good to me

PR #295, opened 2025-07-24
Workaround for Android blank screen

SergeBakharev:
Increase the number of forced window renderings to workaround the Android blank screen issue.


SergeBakharev:
Sounds good. I've made it force updates every 0.25s for the first 5s

Issue #286, opened 2025-07-20
Feature Request: Enable Auto-Vacuum on CA1 via "Ext. Control"


timbot:
Goal:
  Adapt the "Ext. Control" switch to function as a vacuum control for the CA1 model, enabling auto-vacuum functionality.

  Acceptance Criteria:
   1. The feature must be disabled by default.
   2. A user-configurable setting must be available to enable this functionality.
   3. When enabled on a CA1 model, the "Auto Vacuum" switch in the Spindle dropdown should control the "Ext. Control" output.
   4. When enabled, UI elements should refer to this control as "Vacuum" instead of "PowerFan" or "Ext. Control".

  ---

  Development Plan (Proposed)

   1. Add Configuration Setting:
       * Introduce a new boolean setting in `carveracontroller/controller_config.json`.
       * Key: `ca1_ext_control_as_vacuum`
       * Title: Enable Ext. Control as Vacuum (Experimental)
       * Description: For CA1 models, this treats the 'Ext. Control' output as a vacuum and enables the auto-vacuum feature.
       * Default: false

   2. Update UI (`carveracontroller/makera.kv`):
       * Modify the disabled property of the "Auto Vacuum" Switch in the <SpindleDropDown> to be `app.model == 'CA1' and not
         root.ca1_ext_control_as_vacuum`.
       * Update the "PowerFan" label in the <DiagnosePopup> to display "Vacuum" when this feature is enabled: `text: tr._('Vacuum') if
         app.model != 'CA1' or root.ca1_ext_control_as_vacuum else tr._('PowerFan')`.

   3. Update Application Logic (`carveracontroller/main.py`):
       * In the `Makera` class, add a `BooleanProperty` to track the new setting: `ca1_ext_control_as_vacuum = BooleanProperty(False)`.
       * In the `Makera.__init__` method, load the value from the configuration using the appropriate method:
         `self.ca1_ext_control_as_vacuum = Config.getboolean('carvera', 'ca1_ext_control_as_vacuum')`.
       * In the updateStatus method, locate the control_list processing loop. Modify the logic for 'vacuum_mode' to check if the new feature
          is enabled for the CA1 model. If it is, the call should be directed to `self.controller.setExternalControl()` instead of
          `self.controller.setVacuumMode()`.

---

I am willing to implement this feature if it is accepted. Thanks!

timbot:
Thank you, @SergeBakharev !

[...]

PR #320, opened 2025-08-07
Auto-Vacuum on CA1 via "Ext. Control"

SergeBakharev:
This PR adds a new configuration setting to enable the "Ext. Control" switch to function as vacuum control for the CA1 model. enabling auto-vacuum functionality. As discussed in #286

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -2249,6 +2249,10 @@ class Makera(RelativeLayout):
         # model metadata check timer
         Clock.schedule_interval(self.check_model_metadata, 10)
 
+        # Device-specific settings
+        # CA1
+        self.ca1_ext_control_as_vacuum = False
+
         self.has_onscreen_keyboard = False
         if sys.platform == "ios":
             self.has_onscreen_keyboard = True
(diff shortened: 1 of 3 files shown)


carveracontroller/main.py
```diff
@@ -2275,6 +2279,18 @@ def __del__(self):
...
+        # Initialize device-specific settings
+        app = App.get_running_app()
+        if app.model == 'CA1':
+            self.ca1_ext_control_as_vacuum = self.config.getboolean(
+                'Advanced', 'ca1_ext_control_as_vacuum', fallback=False)
+
+        self.spindle_drop_down = SpindleDropDown()
+        self.coord_popup = CoordPopup(self.coord_config)
+
+
```
SergeBakharev, line 2293: This duplicates the existing `setUIForModel()` code. Better to extend the functionality there.

carveracontroller/main.py
```diff
@@ -3900,7 +3916,11 @@ def updateStatus(self, *args):
             elapsed = now - self.control_list['vacuum_mode'][0]
             if elapsed < 2:
                 if elapsed > 0.5:
-                    self.controller.setVacuumMode(self.control_list['vacuum_mode'][1])
+                    logging.info("Setting vacuum mode to: {}".format(self.control_list['vacuum_mode'][1]))
+                    if self.ca1_ext_control_as_vacuum:
+                        self.controller.setExternalControl(100 if self.control_list['vacuum_mode'][1] else 0)
+                    else:
+                        self.controller.setVacuumMode(self.control_list['vacuum_mode'][1])
```
SergeBakharev, line 3923: I see what [you mean](https://github.com/Carvera-Community/Carvera_Controller/issues/286#issuecomment-3162198341). Yes, this has the effect of turning on the vacuum whenever the ca1_ext_control_as_vacuum setting is enabled, not based on the job start/stop.

While tempting it's not possible to send `M851` / `M852` while the machine is playing a gcode file. So doing a conditional on the machine status to send the command isn't the way.

As for
> manually interject the external control m-codes (M851 / M852) before/after the job
This is what I would be tempted to do. It's easy to add before the playback of a file via the `buffer` command ( see how this is used in Makera.play() and Makera.apply()), and running the command immediate after play should queue it to stop after file play back finishes.

However there is potential for edge case issues, for example what happens if you pause gcode playback?

For this reason actually it is probably safest to implement the functionality in the firmware. Basically extend `M331` and `M332` to read a machine config to see if it should use the Ext Control output to perform the vacuum function.

SergeBakharev:
closing as functionality was implemented in https://github.com/Carvera-Community/Carvera_Controller/pull/663

Issue #270, opened 2025-06-30
Macros


jterblanche:
A couple of posts mention the need to:

1.) run custom g-code pre or post job
2.) have additional targets in the goto menu

May I suggest another take on this?

It would be handy to have g-code snippets saved as macros, and have the ability to:
1.) run them manually at any time, from a selectable menu
2.) include them via check boxes (and ordering possibly) in the job run screen, as either pre-job or post-job

An example could be to create a macro that can repeatedly perform a very specific probing sequence with the 3d probe,
and either manually/quickly run it from the macro screen, or include it as a pre-job script to set the origin of repeated jobs.

Hope all this makes sense and thanks heaps for the incredible work so far!

SergeBakharev:
We recently added pendant support (specifically for the WHB04 family of devices) and as part of that we added macro mapping to the buttons of this device.

It would be a cool idea to move the macros out of the Pendant settings, and into it's own settings UI, and then add the ability to map buttons on the control screen to these macros as well as pendant buttons.

I don't have bandwidth myself to pick up this work right now and would want to test to make sure the pendant macro functionality isn't broken. But once I receive my own WHB04 device in a month might have a look at adding this functionality.

Here are some screenshots of the current Pendant macro config screens:

![Image](https://github.com/user-attachments/assets/38e5175d-7c7f-4932-bfed-f281148a29e7)

![Image](https://github.com/user-attachments/assets/54ee2cc2-35b9-4ac4-98eb-bbdbaa5aff0f)

SergeBakharev:
Implemented in #323

PR #323, opened 2025-08-11
Macros

SergeBakharev:
Implements #270 with the addition of 3 macro buttons on the Control screen. These are set in the Controller Settings panel using @yaqwsx's `gcodesnippet` setting type to bring up a gcode text editor and save the macro as json.

The macro name (if set) is used as the text on the button.

Here is how they look when unset:
image

here is how they look if set:
image

If you try to run one without any config:
image

In this PR I've also converted all the `print()` statements to properly use the logger. In future we should add some more log config to the setting screen, right now "debug" is the default level and the logs are written to file per session.

I've tweaked how the spacing between the panel groups in the control panel is done. Previously I was using top/bottom `padding` but this results in unreliable sizing of the grey background shape when the window is resized. I've instead achieved the vertical gaps between the button groups using an empty `Widget` with a prefixed `size_hint_y`.

I've also set a minimum height for these three button groups so that they don't look too weird when the app window is resized. It's not perfect but better than before.

### DIFF
--- a/carveracontroller/makera.kv
+++ b/carveracontroller/makera.kv
@@ -3497,6 +3497,7 @@
                                 orientation: 'vertical'
                                 padding: '3dp'
                                 size_hint_y: 0.3
+                                size_hint_min_y: '41dp'  # 35dp button + 3dp padding top + 3dp padding bottom
                                 BoxLayout:
                                     padding: '3dp'
                                     canvas.before:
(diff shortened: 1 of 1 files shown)


yaqwsx:
I like the changes!

WARIO2412 (approved):
Looks good!
We should check the scaling on small devices after a merge, when the different binaries are easily available

PR #334, opened 2025-08-15
Mini update to use the LEAD position of the step size to scale the feedrate to the wheelspeed

WARIO2412:


SergeBakharev:
Works great for me. 
Only suggestion I have is that I'm worried about the float precision makes the commands look a bit whack but in the distance traveled  and now the feed. I worry that in future this might cause some unexpected issue:
image

carveracontroller/addons/pendant/pendant.py
```diff
@@ -207,7 +206,10 @@ def _handle_jogging(self, daemon: whb04.Daemon, steps: int) -> None:
                         else:
                             self._controller.startContinuousJog(f"{axis}{distance}", None, f"S{daemon.step_size_value}")
             else:
-                self._controller.jog(f"{axis}{distance}")
+                if daemon.step_size == whb04.StepSize.LEAD:
+                    self._controller.jog(f"{axis}{steps * 0.1}", abs(steps * 0.1 / 0.05) * 60 * 0.97)
```
SergeBakharev, line 210: 3 decimal places should be enough right?
```suggestion
                    self._controller.jog(f"{axis}{steps * 0.1}", round(abs(steps * 0.1 / 0.05) * 60 * 0.97), 3)
```

carveracontroller/addons/pendant/pendant.py
```diff
@@ -207,7 +206,10 @@ def _handle_jogging(self, daemon: whb04.Daemon, steps: int) -> None:
...
+                else:
+                    self._controller.jog(f"{axis}{distance}")
```
SergeBakharev, line 212:
```suggestion
                    self._controller.jog(f"{axis}{round(distance, 3)}")
```

carveracontroller/addons/pendant/pendant.py
WARIO2412: Wouldn't it be better to just generally round the output commands to the machine in the jog method ?

PR #341, opened 2025-08-21
Port of new comms protocol from Makera Controller 0.9.12

SergeBakharev:


carveracontroller/main.py
```diff
@@ -4025,7 +4064,20 @@ def updateStatus(self, *args):
...
+                        self.controller.stream.send(b'echo echo\n')
+                        echo = self.controller.stream.getc(10)
+                        if echo == b'echo: echo':
+                            self.message_popup.lb_content.text = tr._('Firmware version mismatch! \nPlease use a Controller with version V0.9.11 or earlier \nto upgrade the firmware to V1.0.4.')
+                            self.message_popup.btn_ok.disabled = False
+                            self.message_popup.open(self)
+                            return
+                        self.echosended = True
+                except:
+                    print(sys.exc_info()[1])
```
SergeBakharev, line 4080: We have better ways of checking the firmware version since we directly have `app.is_community_firmware` and `app.fw_version_digitized`

Please replace for something easier to read that uses those two app attributes.

carveracontroller/main.py
```diff
@@ -3757,6 +3793,7 @@ def doUpload(self, callback):
             #md5 = Utils.md5(self.uploading_file)
             md5 = Utils.md5(displayname)
             self.controller.uploadCommand(os.path.normpath(remotename))
+            time.sleep(0.2)
```
SergeBakharev, line 3796: This is arbitrary. Some time needs to be given for the machine decompress the file but it  shouldn't be a fixed 0.2s.

carveracontroller/main.py
```diff
@@ -3308,7 +3316,12 @@ def check_and_download(self):
     # -----------------------------------------------------------------------
     def download_config_file(self):
         app = App.get_running_app()
-        app.selected_local_filename = os.path.join(self.temp_dir, 'config.txt')
+        if kivy_platform == 'ios':
+            import tempfile
+            tmp_file = tempfile.gettempdir()
+            app.selected_local_filename = os.path.join(tmp_file, 'config.txt')
+        else:  # inserted
+            app.selected_local_filename = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'config.txt')
```
SergeBakharev, line 3324: Same as https://github.com/Carvera-Community/Carvera_Controller/pull/341/commits/cda588739b408d902f1a7428bdcaea7a8ed4af25#r2289723789

carveracontroller/main.py
```diff
@@ -3294,7 +3297,12 @@ def check_and_download(self):
         remote_path = self.file_popup.remote_rv.curr_selected_file
         remote_size = self.file_popup.remote_rv.curr_selected_filesize
         remote_post_path = remote_path.replace('/sd/', '').replace('\\sd\\', '')
-        local_path = os.path.join(self.temp_dir, remote_post_path)
+        if kivy_platform == 'ios':
+            from os.path import expanduser
+            local_path = join(expanduser('~'), 'Documents')
+            local_path = join(local_path, remote_post_path)
+        else:  # inserted
+            local_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), remote_post_path)
```
SergeBakharev, line 3305: Should use our version `local_path = os.path.join(self.temp_dir, remote_post_path)`

carveracontroller/main.py
```diff
@@ -2129,6 +2129,9 @@ class Makera(RelativeLayout):
     fw_version_checking = False
     fw_version_checked = False
 
+    echosended = False
+    echosending = False
```
SergeBakharev, line 2133: This is used for some reason by makera as a way of version checking. Per https://github.com/Carvera-Community/Carvera_Controller/pull/341/commits/cda588739b408d902f1a7428bdcaea7a8ed4af25#r2289714852 We have a better way. Please remove it.

carveracontroller/XMODEM.py
```diff
@@ -823,34 +492,29 @@ def putc(data, timeout=3):
                 size = len(data)
             else:
                 size = None
-
-            print(('putc(', repr(data), repr(size), ')'))
+            print(("putc(", repr(data), repr(size), ")"))
```
SergeBakharev, line 495: No print statements for debugging please. Use logger.debug()

carveracontroller/XMODEM.py
```diff
@@ -738,81 +432,56 @@ def _putc(data, timeout=timeout):
...
-        print(('so', so))
         def getc(size, timeout=3):
             read_ready, _, _ = select.select([so], [], [], timeout)
             if read_ready:
                 data = so.read(size)
             else:
                 data = None
-
-            print(('getc(', repr(data), ')'))
+            print(("getc(", repr(data), ")"))
```
SergeBakharev, line 484: https://github.com/Carvera-Community/Carvera_Controller/pull/341/commits/cda588739b408d902f1a7428bdcaea7a8ed4af25#r2289728438

carveracontroller/XMODEM.py
```diff
@@ -1,718 +1,412 @@
...
-
-
-"""
 from __future__ import division, print_function
-
-__author__ = 'Wijnand Modderman <maze@pyth0n.org>'
-__copyright__ = ['Copyright (c) 2010 Wijnand Modderman',
-                 'Copyright (c) 1981 Chuck Forsberg']
-__license__ = 'MIT'
-__version__ = '0.4.5'
```
SergeBakharev, line 95: Let's not remove the acknowledgement to the original author

carveracontroller/WIFIStream.py
```diff
@@ -164,13 +166,13 @@ def putc(self, data, timeout = 0.5):
...
         stream = open(filename, 'rb')
-        result = self.modem.send(stream, md5 = local_md5, retry = 10, callback = callback)
+        result = self.modem.send(stream, md5=local_md5, retry=50, callback=callback)
         stream.close()
         return result
 
     def download(self, filename, local_md5, callback):
         stream = open(filename, 'wb')
-        result = self.modem.recv(stream, md5 = local_md5, retry = 10, callback = callback)
+        result = self.modem.recv(stream, md5=local_md5, retry=50, callback=callback)
```
SergeBakharev, line 175: ... 50 retries seems excessive. Users will think the app is broken if it spends this much time retrying. Let's continue using the original value of 10, or maybe even reduce it to 3.

carveracontroller/WIFIStream.py
```diff
@@ -164,13 +166,13 @@ def putc(self, data, timeout = 0.5):
     def upload(self, filename, local_md5, callback):
         # do upload
         stream = open(filename, 'rb')
-        result = self.modem.send(stream, md5 = local_md5, retry = 10, callback = callback)
+        result = self.modem.send(stream, md5=local_md5, retry=50, callback=callback)
```
SergeBakharev, line 169: https://github.com/Carvera-Community/Carvera_Controller/pull/341/commits/cda588739b408d902f1a7428bdcaea7a8ed4af25#r2289733565

carveracontroller/WIFIStream.py
```diff
@@ -3,27 +3,30 @@
...
 class MachineDetector:
+
     def __init__(self):
         self.machine_list = []
         self.machine_name_list = []
         self.sock = None
         self.t = None
         self.tr = None
-
+        return
```
SergeBakharev, line 28: init methods does not explicitly return a value. Its purpose is to initialize the attributes of a newly created object, not to produce a return value for assignment.

carveracontroller/WIFIStream.py
```diff
@@ -98,17 +97,20 @@ def recv(self):
...
-        ip_port = address.split(':')
-        self.socket.settimeout(2)
-        self.socket.connect((address.split(':')[0], (int)(address.split(':')[1]) if len(ip_port) > 1 else TCP_PORT))
-        self.socket.settimeout(SOCKET_TIMEOUT)
-
-        return True
+        try:
+            self.socket = socket.socket(family=socket.AF_INET, type=socket.SOCK_STREAM)
+            ip_port = address.split(':')
+            self.socket.settimeout(3)
```
SergeBakharev, line 103: Eh? Whats the point of defining `SOCKET_TIMEOUT`  as 5 if it's not even used.

SergeBakharev (commented):
There is a lot of iOS related changes here, that I'm not comfortable commenting on. I don't know if they conflict with the existing stuff written by @zittix or not. I don't like the idea of taking in code changes from Makera without verifying them, so I would prefer to drop all the iOS changes.

I see some re-occuring issues:
- Makera is using some inane way of testing the firmware version for some reason with `controller.echosending` We shouldn't propagate that when we already do better with our own `MakeraApp.is_community_firmware` and `MakeraApp.fw_version_digitized`
- We use a temp_directory for temporary files but Makera is saving the files directly next to the controller executable. We should continue to use `MakerApp.temp_dir`

A point of discussion: This change is breaking, are we okay with that? Previous versions of firmware will not be supported. Which is not ideal but I'm okay with. We should however make it clear in our messaging the order that the updates should be applied: First new firmware uploaded with previous controller version before using the new controller.

Sorry I wrote my previous comments while looking at a specific commit so I wasn't able to directly suggest code changes.

PR #344, opened 2025-08-26
Pushing Cancel on the Changing Tool popup stops g-code playback

SergeBakharev:
Pushing Cancel on the Changing Tool popup stops g-code playback


WARIO2412:
self.controller.stopRun is actually the wrong method to call here. It would need something like self.controller.abortCommand.
Unfortunately. This needs changes to the firmware to work. I changed it to issue the abort command, but the machine stays in the manual toolchange state.

SergeBakharev:
Ah, I see. Bummer

PR #403, opened 2025-10-06
Respond to double clicks in local/remote file browsers

acj:
A common pattern in UI file chooser dialogs is that a double click runs the default action on the selected file. The Controller's dialogs could do the same, e.g. by loading the file in the remote dialog and uploading (+ selecting) in the local dialog. This PR wires up both.

Tested on macOS. I'll test on linux tonight or tomorrow, too.


acj:
Thanks. I can't merge, but this looks good on macOS and linux.

PR #406, opened 2025-10-14
Clean up typos

acj:
As I've come across typos and other non-functional fixes that didn't seem worth rolling into an unrelated PR, I've kept them in a branch. Here's a batch of them.


SergeBakharev:
Nice, thanks @acj

PR #417, opened 2025-11-20
Add workflow for backing up the machine's configuration files

### LINKED ISSUE #97: Add option to backup config.txt
The Carvera requires only a single file from the sd card to boot: config.txt

This file is unique per machine because at the factory they tweak the anchor 1 offsets individually.

Makera keep a copy of each machine config file before it leave the factory but it would be nice anyway to have an option in the Controller to backup this small text file. Probably would make sense to put it on the firmware upgrade screen.

acj:
This PR adds plumbing and UI for downloading a backup copy of the machine's various config files in case the machine corrupts them, or the SD card dies, or some other bad thing happens. It operates from a hard-coded list of known config files and should be easy to extend in the future.

Tested on macOS and Linux. I'm also going to test on iOS since the file chooser is a bit different.

I'm interested in general feedback, and specifically on these points:
1. Are we okay with using the current directory (I think this is always the directory where the app is launched from) as the default for where to save these config files? The user's home directory might also make sense but will vary across platforms.
2. How does the interaction in the settings panel feel?
3. Now that the backup option is in the settings panel, do we want to leave it on the firmware upgrade screen too?
4. Does this work on Windows?

Resolves #97.

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -4887,7 +4887,8 @@ class Makera(RelativeLayout):
             self.config_popup.settings_panel.add_json_panel('Machine - Basic', self.config, data=json.dumps(basic_config))
             self.config_popup.settings_panel.add_json_panel('Machine - Advanced', self.config, data=json.dumps(advanced_config))
             self.config_popup.settings_panel.add_json_panel('Machine - Restore', self.config, data=json.dumps(restore_config))
-            self.config_popup.settings_panel.add_json_panel('Machine - Backup', self.config, data=json.dumps(backup_config))
+            if kivy_platform not in ['android', 'ios']:
+                self.config_popup.settings_panel.add_json_panel('Machine - Backup', self.config, data=json.dumps(backup_config))
         return True
 
     # -----------------------------------------------------------------------
--- a/carveracontroller/makera.kv
+++ b/carveracontroller/makera.kv
@@ -2222,6 +2222,7 @@
                                 size_hint_x: None
                                 width: '180dp'
                                 text: tr._("Back up machine config")
+                                disabled: kivy.utils.platform in ["android", "ios"]
                                 on_release:
                                     app.root.start_back_up_config()
                             Button:


carveracontroller/main.py
```diff
@@ -5,6 +5,8 @@
 # import os
 # os.environ["KIVY_METRICS_DENSITY"] = '1'
 
+CONFIG_FILES_TO_BACK_UP = ['/sd/config.txt', '/sd/custom_tool_slots.txt', '/sd/flex_compensation.dat']
```
acj: Should I include any other files? Maybe `config.default` as well?

carveracontroller/makera.kv
```diff
@@ -2218,6 +2218,12 @@
                                 halign: "right"
                                 valign: "middle"
                                 text_size: self.size
+                            Button:
+                                size_hint_x: None
+                                width: '150dp'
+                                text: tr._("Back up configuration")
```
acj: I'd appreciate any wordsmithing advice here. I couldn't think of another short _and_ clear way of describing this that would fit on a button.

carveracontroller/XMODEM.py
```diff
@@ -444,8 +444,8 @@ def callback(success_count, error_count)
         while True:
             # first try CRC mode, if this fails,
             # fall back to checksum mode
-            if error_count >= retry:
-                self.log.info('error_count reached %d, aborting.', retry)
+            if error_count >= 50:
+                self.log.error('error_count reached %d, aborting.', retry)
```
acj: This is related to the `time.sleep` call in `download_config_files`. If I don't include it, something in the communication with the machine jams and spits out these error_count messages, which easily got lost in the log noise.

When I added logging in here, I saw that the comm buffer had "Info: canceled by remote!" in it.

carveracontroller/main.py
```diff
@@ -3455,26 +3530,28 @@ def finishLoadConfig(self, success, *args):
         self.updateStatus()
 
     # -----------------------------------------------------------------------
-    def doDownload(self):
+    def doDownload(self, remote_path, local_path, show_progress=True):
```
acj: Passing these paths explicitly was also an unplanned change, but it made the download flow much easier to reason about. Fewer global variables being reused for different things, etc.

carveracontroller/main.py
```diff
@@ -3403,19 +3424,73 @@ def check_and_download(self):
...
+        Clock.schedule_once(partial(self.progressStart, tr._('Downloading config files...'), None), 0)
+
+        self.fill_remote_dir_callback = self.download_config_files
+        self.file_popup.remote_rv.list_dir("/sd")
+
+    # -----------------------------------------------------------------------
+    def download_config_files(self, remote_paths):
+        matching_paths = []
+        for file_info in remote_paths:
+            if file_info['path'] in CONFIG_FILES_TO_BACK_UP:
```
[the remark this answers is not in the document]
acj: @SergeBakharev Sorry, rereading this I realized that I didn't quite answer your question. The paths in `CONFIG_FILES_TO_BACK_UP` are okay as-is. The bug on Windows (fixed in e1721a69f4d126fb13d9881a54db3d665d76a196) was from using `os.path.join` in a place where it didn't make sense -- we were joining `/sd` with a remote filename with `\\` as the platform-specific delimiter and then trying to use that garbled path later. As you said, we might as well keep these paths in the format that the firmware expects.

There are several string replacements in the Makera code [like this one](https://github.com/acj/Carvera_Controller/blob/7109c8eb8fbffd6e3340237b408011809cf5a67d/carveracontroller/main.py#L3424) that were probably masking the bug. So far I haven't found any new bugs from my fix, but that's something to watch for.

carveracontroller/config_c1.json
```diff
@@ -623,5 +623,13 @@
...
         "section": "Restore",
         "key": "default"
+    },
+    {
+        "type": "options",
+        "title": "Back up configuration files",
+        "desc": "config.txt, flex_compensation.dat, etc. will be backed up to your computer",
+        "section": "Backup",
+        "key": "backup",
+        "options": ["Back up files now"]
```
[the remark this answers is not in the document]
acj: Thanks! I used a similar approach in https://github.com/Carvera-Community/Carvera_Controller/pull/417/commits/de3e3bac61044cd65e0738e35fb0b0d0633fd3e6.

carveracontroller/main.py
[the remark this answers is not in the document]
SergeBakharev: Ah, a rough `os.path.join` makes sense! Thanks for the reply.

acj:
@SergeBakharev @WARIO2412 Okay, I think I've addressed all of your feedback and this is ready for another review/test. Thanks again.

SergeBakharev:
Works amazingly, great work! Tested on Windows and Linux (Pop_OS!)

PR #424, opened 2025-11-27
Automatically connect over wifi on startup if it's configured

### LINKED ISSUE #421: Enhancement: Auto connect on launch
Ideally if the reconnect option is available, when the controller launches it could attempt to auto reconnect, probably failing silently if the machine is not available.

acj:
This works reliably for me on macOS, Linux, and in the iOS simulator.

Resolves #421


SergeBakharev:
Nice one!

PR #448, opened 2025-12-14
Fix cleanup handler when app exits

acj:
We try to save the last-used window size on exit, but that hasn't been working for me. I traced it to the `__del__` function not being called, and indeed the [Python docs](https://docs.python.org/3/reference/datamodel.html#object.__del__) say that it's not guaranteed to be called. Kivy's docs, OTOH, recommend [on_request_close](https://kivy.org/doc/stable/api-kivy.core.window.html#kivy.core.window.WindowBase.on_request_close) as the hook for this, so this PR wires up that hook as our cleanup/exit handler.

Tested on macOS and Linux, with and without my pendant connected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Enhancement: Select a file when it's double clicked in the remote file browser
 - Enhancement: Automatically connect to the machine on startup if its wifi address is configured
 - Enhancement: CI workflow for building iOS app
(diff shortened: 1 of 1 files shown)


acj:
@SergeBakharev The changelog still has unreleased entries that shipped in 2.0.0. Do you want me to sync it with what's on the main branch?

PR #452, opened 2025-12-17
Support sending multiple MDI commands at once

acj:
A common feature on industrial CNC machines is sending multiple commands at once. They're often separated by semicolons, but because Smoothieware uses semicolons as a comment indicator, we need to handle it differently. This PR makes the MDI input box slightly taller to accommodate more than one command. It also changes the default enter key behavior to insert a newline. To send the commands to the machine, use control+enter.

Original description:
~A common feature on industrial CNC machines is sending multiple commands at once, separated by a semicolon, e.g. `G0 X25; Y30`. They're executed sequentially. We can't easily support the abbreviated syntax (second `G0` is implied) without a parser, but I think a simpler form of multi-command is still useful: `G0 X25; G0 Y30`.~

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -5316,7 +5316,10 @@ class Makera(RelativeLayout):
             if to_send.lower() == "clear":
                 self.manual_rv.data = []
             else:
-                self.controller.executeCommand(to_send)
+                sanitized_to_send = '\n'.join([line for line in to_send.split('\n') if line.strip().lower() != "clear"])
+                if sanitized_to_send != to_send:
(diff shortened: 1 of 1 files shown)


faecorrigan:
bringing in some discussion from discord:

It would be good for the MDI to accept Enter and Shift enter to create newlines. Enter would be the preferred newline method and would be what is in the documentation.


and then run a command on control + enter. That gates running a command behind a simultaneous keypress

acj:
image

faecorrigan:
@acj 
This looks clean, I think your instinct to wrap the clear function in a handler is a good idea. The error shouldn't cause any harm on the machine, but it would be cleaner to handle it controller side.

Once that is done I am comfortable merging it to dev

PR #455, opened 2025-12-19
Prevent keyboard jog when MDI text box has focus

acj:
I stumbled onto this bug while testing the other MDI tweaks. If you enable keyboard jogging and then use the arrow keys while you're editing a command in the MDI text box, you'll cause a machine move that's probably unexpected and could be dangerous.


faecorrigan:
@acj a request: can you add changelog information into the changelog.md file as you go with these small pushes?
otherwise the two open ones look good, I might be able to test them and push them into dev

acj:
Sorry, I should have mentioned that. We need to land #449 so that the changelog has the 2.0.0 changes in the right section. Otherwise there are going to be messy merge conflicts.

PR #465, opened 2026-01-03
Disable and hide irrelevant buttons when browsing for firmware file to upload

### LINKED ISSUE #460: Firmware file browser shows Download Button
When browsing firmware for upload the file browser shows a number of irrelevant and confusing buttons:
* Download
* Upload and Select

These should be removed.

Image

acj:
Resolves #460. I think the logic in `update_local_buttons` is correct, but please check me @SergeBakharev.


SergeBakharev:
Perfect!

PR #490, opened 2026-02-01
Restore the previously-loaded background image in Config and Run

### LINKED ISSUE #422: Enhancement: Save and restore background image from job start page
Background fixture plate image should persist through application restart

acj:
Resolves #422


carveracontroller/main.py
```diff
@@ -1653,9 +1653,10 @@ def load_config(self, config):
         self.config = config
 
     def update_background_image(self, new_source):
-        if self.bg_rect and new_source != "None":
-            self.bg_rect.source = new_source
+        if new_source != "None":
             self.bg_image = new_source
+            if self.bg_rect:
```
acj: Without this change, the background image wouldn't actually load even though it's selected in the spinner widget.

PR #502, opened 2026-02-08
keyboard and pendant jogging in probing screen

faecorrigan:
This PR adds the keyboard and pendant jogging buttons to the probing screen. 

When opening the probing screen or clicking into any of the text input fields on the probing screen, keyboard jogging is disabled to prevent mishaps. 

I had to restructure how some of the text input is displayed on the keyboard and pendant jogging buttons

image


SergeBakharev:
My pendent is also broken in this branch, no issues in the latest develop though (commit be3f8a15ae759ecfb62e4976dd5c5f600d434f11)

PR #518, opened 2026-02-17
Pr add max rpm config

faecorrigan:
This one is pretty simple, it adds a new config variable to the controller that limits maximum rpm on the community firmware version of the machine. 
Also in this PR is some code to prevent disconnect if a config setting is present on the controller side but not the machine side


faecorrigan:
to be used with https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/249

PR #559, opened 2026-04-10
Display alarm message in halt popup

Lyrkan:
Screenshot_2026-04-10_23-01-16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 [unreleased]
-
+- Enhancement: Display alarm message in halt popup
 
 [2.1.0]
 - Enhancement: Add right-click menu option to clear resume-at-line setting
--- a/carveracontroller/CNC.py
+++ b/carveracontroller/CNC.py
@@ -101,6 +101,7 @@ class CNC:
 
         'wpvoltage'  : 0.0,
         'halt_reason': 1,
+        'alarm_message': '',
 
         "planner"    : 0,
         "rxbytes"    : 0,
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -1443,6 +1443,7 @@ class Controller:
             #self.stream.send(b"\n")
             self._gcount = 0
             self._alarm = True
+            CNC.vars["alarm_message"] = ''
             # Reset manual disconnect flag when connection is established
             self._manual_disconnect = False
             try:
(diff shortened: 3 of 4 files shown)


SergeBakharev:
I suggest we discuss this PR further next Community projects meeting. The reason the existing halt system uses numbers is so that the language specific message can be shown. By directly passing the alarm message we don't get the opportunity to translate it to the user's language.

I think if a halt code is provided we show the prepared translated text, and in addition append the English alarm text is a decent compromise. 

Our commitment to international language support has been eroding with every release, so there is a discussion to be had acknowledging that and agreeing on a reasonable way forward.

Lyrkan:
The halt code is already translated, the issue is that some of them (like "Halt manually" here, or "Probe Fail") are not that explicit/regroup multiple errors and require users to check the MDI to know what really happened.

Translating the alarm messages as well would be good but some have variables inside of them so it won't be as easy as just mapping an error code to a translation.

Unrelated but I just noticed that some messages are also not formatted properly in the firmware, so they won't be picked up by my changes :
* `Alarm:<message>` instead of the more common `ALARM: <message>`. I think it'd probably be better to normalize them in the firmware but I can also just change the detection here
* [No "ALARM:" prefix at all](https://github.com/Carvera-Community/Carvera_Community_Firmware/blob/feab653def96a959e695049a4f828aeddbcd5547/src/modules/tools/endstops/Endstops.cpp#L596), not sure if that's intended since we have one just below 🤔

SergeBakharev:
Right, Alarm messages != Halt codes. I don't think we practically can do better than you have described without implementing some sort of alarm message structure. Merging!

As for the inconsistency in the firmware, we should indeed normalize them.

faecorrigan:
I am auditing the firmware and changing all halts to format with ALARM. 
In the firmware there are several variants of the alarm including HALTED: and ERROR: 
for consistency I will rename those to ALARM:

faecorrigan:
@Lyrkan @SergeBakharev 
A thought: Would it make more sense to change the code to ERROR: instead of ALARM: ?

Lyrkan:
> @Lyrkan @SergeBakharev I thought: Would it make more sense to change the code to ERROR: instead of ALARM: ?

I wouldn't mind "ERROR:", I even used it by mistake in my last PR because I misremembered the format...

SergeBakharev:
I also like ERROR over ALARM. Sounds more generic, while some text with ALARM can be confusing.

Issue #565, opened 2026-04-21
Add popup notice when using stock firmware instead of the Community Firmware


software-2:
I see on Discord there's a number of people who have problems because they aren't running the Community Firmware, but they are using the Community Controller. Since it's basically required to use the Community Firmware, why not add a helpful nag popup when it's detected the connected machine is running stock?

<img width="1606" height="938" alt="Image" src="https://github.com/user-attachments/assets/76a183db-2873-4d45-b613-8ab3cc3b59d4" />

PR #566, opened 2026-04-21
Add popup notice when using stock firmware instead of the Community Firmware

software-2:
Implement #565


SergeBakharev:
@faecorrigan What do you think about this? Too opinionated? Increasingly the firmware is diverging, so it's not untrue comment to say that we recommend using Community Firmware with the Community Controller.

Maybe the language could be softer along the lines of "Complete functionality available when using both Community Firmware and Community Controller"

faecorrigan:
I think a softer approach is better
I haven't compiled it to run, but does it pop up every time you open the controller with old firmware?
If so, that is a lot and it should be gated behind an ignore future messages

software-2:
Yes, as-is, it would nag you each time. I did say "STRONGLY". :D 

I can soften the language and add a "don't show again" option.

SergeBakharev:
softer language + don't show again would be perfect. I would also use https://carvera-community.gitbook.io/docs/compatibility as the url. The goal was to list any known compatibility issues on that page as we learn about them.

PR #574, opened 2026-04-27
Fix RecycleView crashes by updating data on main thread

Lyrkan:
`monitorSerial` does not run on the main Kivy thread  which means that directly trying to update widgets from there could crash the app with that kind of stacktrace:

image

A similar issue was present in `execCallback` but hidden by a try/catch block.

Instead of doing that I added an `_append_to_mdi` function annotated with the `mainthread` decorator as recommended on this page: https://github.com/kivy/kivy/wiki/Working-with-Python-threads-inside-a-Kivy-application


SergeBakharev:
Thank you for looking into how to do this properly!

acj:
Related: #409, #454

I agree that this is the way to go - being strict about updating the underlying array and working with UI widgets from the main thread is the only safe method. The Makera code wasn't disciplined about threads, so I expect that we'll run into more (and possibly new) bugs. But it's worth a try imo. I'll try this on macOS to see if the weird MDI bouncing behavior comes back.

There are a lot of UI-related functions that should probably use the `@mainthread` annotation for safety, but one thing at a time. 😄

PR #577, opened 2026-05-01
Only re-render the G-Code viewer scene if something has changed

Lyrkan:
Some straightforward changes to avoid re-rendering the scene at 60 FPS when no update is needed by returning early in the render callback.

On my laptop, CPU usage dropped from a constant 40% after loading a file to about 4% at idle

I also renamed the `increase_angle` callback to `_on_frame_tick` since I can't see any valid reason why it was called this way...


SergeBakharev:
Massive!

PR #578, opened 2026-05-01
Added linting and formatting tooling

software-2:
More prep for refactoring the monolith. The goal here is to enforce some code quality standards moving forward. As such, ruff is disabled on existing files. (I would prefer to fix the formatting and real bugs that are identified bit by bit, instead of trying to tackle main.py.)


software-2:
There's definitely some opinionated configuration options here, none of which I am strongly attached to (except enforcing a separation between Kivy and machine operations... that one I care about!).

@SergeBakharev @faecorrigan If either of you have any thoughts/opinions about what rules to enforce, please let me know. I'm more than happy to change things around. I'd just like to roughly get the standards set before I start ripping things out of main.py.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - id: ruff-format
+
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy (strict on /machine/, /ui/)
+        entry: poetry run mypy carveracontroller
+        language: system
+        types: [python]
+        pass_filenames: false
```
SergeBakharev, line 24: I think it will be too hard to avoid pulling in all the past changes into type checking. I suspect this won't end well

SergeBakharev:
Up for discussion but I think we should also run the linting via CICD as a gate for PR merge

.pre-commit-config.yaml
software-2: There's an exclude list at https://github.com/Carvera-Community/Carvera_Controller/pull/578/changes/7035119a7237d98ceb5ad276465cd0ccd7eb5c7b#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R123

What scenario(s) do you suspect might cause problems here?

software-2:
> Up for discussion but I think we should also run the linting via CICD as a gate for PR merge

Oh, I agree, most definitely. I don't have access to test that functionality though.

.pre-commit-config.yaml
SergeBakharev: My understanding is that any code referenced in the analyzed files that comes from excluded files is treated as returning/accepting `Any`. 

So if excluded.py defines `def foo() -> int` and you import and call `foo()` in an analyzed file, mypy will treat the return type as `Any` rather than `int`. 

I've seen this in other code bases where the `Any`s propagate deeply into the analysed files and silently undermines type checking and in turn trust in the system.

I also note that large amounts of the Kivy core is written in Cypthon and thus has no type hints.

I'm happy to merge this PR and give it ago, but I don't expect miracles.

software-2: Ah, I understand what you mean now, thanks for explaining. 

From that perspective, it's certainly not ideal, but still the lesser of two evils, IMO. My goal is to eventually reduce those files into a more manageable size so that the ignores can be removed. And I'm sure at that point there'll be more type fixing to do. The joys of refactors.

PR #583, opened 2026-05-05
Settings: Fix Apply button not persisting changes and re-enabling properly

### LINKED ISSUE #576: Apply button does not persist settings between restarts
While working on #575 (from the latest commit of `develop`) I noticed that the "Apply" button of the settings screen has some incoherent behaviors that may have been introduced by #564.

For instance:
* if I change a setting (e.g. disabling "Show tooltips"), hit "Apply", then "Close" the settings is applied correctly to the current instance but doesn't seem to be persisted when I restart the app
* however, if I change that same setting but instead directly hit "Close" and then "Confirm" in the new "Unapplied Changes" modal, then the new value is kept after restarting the app

The enabled/disabled state of that button is also acting weird. Let's say the "Show tooltips" option is initially disabled:
* When opening the settings screen the "Apply" button is disabled (OK)
* After toggling on the option the "Apply" button becomes enabled (OK)
* After clicking on "Apply" the changes are applied and the button becomes disabled again (OK)
* Without closing the settings screen if I then toggle off the option again the button stays disabled (should not be the case)

Lyrkan:
Fixes #576

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -1190,7 +1190,7 @@ class ConfigPopup(ModalView):
             makera.confirm_popup.open()
             return True  # cancel the dismiss
 
-    def _apply_and_close(self):
+    def _apply_changes(self):
         app = App.get_running_app()
         # Write pending widget values to their Config instances
         for widget in self._all_setting_items():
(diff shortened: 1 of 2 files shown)


software-2:
Hah, just as soon as I comment. Nice catch!

PR #585, opened 2026-05-07
Pendant: Add gamepad support

Lyrkan:
Adds gamepad support as a new pendant type.
Not as good as a WHB04 but still usable and people are more likely to have one lying around :)

Some notes about it:
* it does support both step/continuous mode but since there is no feedback about the selected % on the controller I added a text on the "Pendant jogging" button
* regarding the continuous mode, I initially wanted to implement an optional way to change the speed based on the inclination of the stick but didn't because the firmware doesn't allow to change the speed of an active jog without stopping it first (which resulted in a really bad user experience).
* I've changed the Pendant settings menu so it only shows settings relevant for the current pendant
* I've tested it on Linux with a bluetooth 8bitdo SN30 pro, but the code should work with all controllers supported by SDL2 since that's what Kivy uses internally. I'll do some more tests on Linux/Windows but I'd really like some feedback on other configs (~I'll try to setup the Github workflow on my fork to make it easier~ __EDIT: [Binaries for Windows/Mac/Linux are available here](https://github.com/Lyrkan/Carvera_Controller/releases/tag/untagged-59d5f69c328b0d20b369)__)

image

image

image

image


SergeBakharev:
Awesome, love your work! I'll review over the weekend and test on different platforms that I have available to me. I suspect iOS/Android are their own special case

Lyrkan:
Added a new commit to handle hot-plugging a controller since [Kivy only retrieves connected joysticks during intialization](https://github.com/kivy/kivy/blob/2.3.1/kivy/core/window/_window_sdl2.pyx#L261-L264C13)...

There isn't a pretty way to do that so the code is a bit hacky, but worst case it should still work if the controller is connected before starting the app. I tested it on Windows and Linux and it seemed to work quite well, but I'm more worried about Mac/iOS/Android 😅.

Also a side-note regarding the default bindings: since Kivy uses [`SDL_Joystick`](https://wiki.libsdl.org/SDL2/CategoryJoystick) and not[ `SDL_GameController`](https://wiki.libsdl.org/SDL2/CategoryGameController) the bindings are done on internal button/stick/... IDs and not high-level mappings ([axes](https://wiki.libsdl.org/SDL2/SDL_GameControllerAxis)/[buttons](https://wiki.libsdl.org/SDL2/SDL_GameControllerButton)). This means that between OSes or even different versions of a given controller the IDs can change and the default mappings may not work all the time. I used https://github.com/mdqinc/SDL_GameControllerDB as a reference for the ones used in this PR.

PR #587, opened 2026-05-08
Monitoring graphs

SergeBakharev:
This adds a spindle monitoring graph to the gcode viewer.

Similar to the Camera screen the graph is hidden behind a collapsed splitter.

If a camera is detected the splitter shares the horizontal bar with both a Camera and Graph show/hide button:
Screenshot 2026-08-29 at 3 50 08 pm

Showing one has it expand the splitter and use the whole real estate:

Screenshot 2026-08-29 at 3 51 45 pm

Expanding both has them share the space with a vertical splitter:
Screenshot 2026-08-29 at 3 52 31 pm

The Graph settings are as per below:
Screenshot 2026-08-29 at 3 53 05 pm

With the available metrics being:
* None
* Temp
* Spindle Power
* Spindle RPM
* Feed

I'm not sure how best to describe the Spindle Power metric. I added in back in 2.1.0 https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/195 It actually shows the spindle power being requested from the motor controller. Due to the speed the firmware calculates the request though, it's effectively the spindle power. I'm not sure if "Spindle Power Request %" is more or less confusing than "Spindle Power".

In future, I would like to add the ability for user extension for metrics. On my own machine I've been interfacing directly with the motor controller to gather it's view on the spindle power %. This could easily be a way for people to develop integration with their own solutions.


SergeBakharev:
Oh dang it, I think my understanding of the LPC <-> spindle motor controller is wrong. Spindle power metric doesn't actually work. 

The LPC doesn't command torque, it's commanding speed but doing so without knowing the voltage scaling on pwm, that's why it reads the rpm feedback and adjusts the pwm up/down

SergeBakharev:
It's shame because this UI is really slick, but I'm going to shelve this graphing PR. Without a Spindle Power source metric it's pointless for most people. It was only working for me because I had wiring in a different spindle motor controller.

PR #588, opened 2026-05-09
Fix: Disable trackpad being treated as touchscreen on Linux

### LINKED ISSUE #571: Linux laptop trackpad being treated as touch screen by Kivy
I'm running the Carvera-Community controller app on my Linux Framework 13 laptop (Debian 13, KDE with Wayland). When you run the app, if you try to move the mouse cursor around to select files to load, it starts generating events to scroll file lists, or moving the tool head around the machine. It seems like Kivy is treating the trackpad on my laptop as a touch screen.

I've found some other articles discussing this problem and a solution involving editing ~/.kivy/config.ini, but this isn't very easy to maintain and you have to run the app once to generate the config file to edit:

https://stackoverflow.com/questions/60890388/kivy-disable-touchpad-touchscreen-like-behavior
"commented out the probesysfs line in the kivy config (~/.kivy/config.ini)"

https://stackoverflow.com/questions/74646535/kivy-laptop-touch-pad-mouse-cursor-move
comment out probesysfs line in ~/.kivy/config.ini:

[input]
mouse = mouse
#%(name)s = probesysfs

Has anyone else seen this behavior and have a solution for it? I remember when working with Godot a few years ago, there were some trackpad -> touchscreen issues that were specific to certain trackpad controllers.

waynepiekarski:
Fixes #571 

Problem
-------
On Linux, Kivy's default configuration adds a catch-all input provider entry:

    %(name)s = probesysfs,provider=hidinput

This causes Kivy to register ALL HID input devices — including laptop trackpads — as touch/multitouch screens. The result is that moving the trackpad generates spurious touch events, making the UI behave as if it's a touchscreen and causing erratic, broken interaction on normal laptops.

Fix
---
In set_config_defaults() in main.py, when running on Linux:

1. Remove the default '%(name)s' probesysfs/hidinput entry before it registers the trackpad as a touch device.

2. Re-add probesysfs using the mtdev backend (the Linux multitouch library, designed for real touchscreen hardware) with a case-insensitive name filter:

       touchscreen_%(name)s = probesysfs,provider=mtdev,match=(?i)touchscreen

   This only registers devices whose kernel name contains "touchscreen" (e.g. "ELAN Touchscreen", "Wacom Touchscreen"). Trackpad device names never contain this string, so they are excluded.

This fix is applied programmatically at startup — no manual editing of ~/.kivy/config.ini is required.

Behaviour after fix
-------------------
- Laptop trackpad: works as a normal mouse cursor (no touch events)
- Real touchscreen hardware: continues to work via mtdev
- Android / iOS: unaffected (the fix is Linux-only)
- macOS / Windows: unaffected (probesysfs is a Linux-only provider)

Tested on
-------------------
Debian 13 Stable with these laptops:
- Framework 13 (trackpad-only)
- Chuwi Minibook X (with trackpad and touch screen)
- GPD Win Max 2 (with trackpad and touch screen)
- Dell XPS 13 (trackpad-only)
Verified the problem existed before this diff, and checked that this change fixes everything with no new regressions. Check that existing trackpads operate as a mouse as expected, and a real touch screen continues to operate as a touch screen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fixed: Spindle temp reporting when running Analog type spindle without rpm reporting
 - Fixed: Fit the gcode viewer to the path's bounding box instead of its max X/Y/Z
 - Fixed: Last character of the current file was sometimes missing in the file viewer
+- Fixed: Disable trackpad being treated as touchscreen on Linux
 
 [2.1.0]
 - Enhancement: Add right-click menu option to clear resume-at-line setting
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -6620,6 +6620,17 @@ def set_config_defaults(default_lang):
     if not kivy_platform in ['android', 'ios']:
         Config.set('input', 'mouse', "mouse,multitouch_on_demand") # disable multitouch simulation on non-mobile platforms
 
+    if kivy_platform == 'linux':
+        # Remove the default probesysfs entry that treats trackpads as touchscreens.
+        # Kivy's default adds '%(name)s = probesysfs,provider=hidinput' which picks up
+        # all HID devices including laptop trackpads.
+        if Config.has_option('input', '%(name)s'):
+            Config.remove_option('input', '%(name)s')
+        # Re-add probesysfs using mtdev, filtered to devices whose name contains
+        # "touchscreen" (case-insensitive). This preserves real touchscreen support
+        # while excluding trackpads, which never have "touchscreen" in their device name.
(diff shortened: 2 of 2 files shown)


SergeBakharev:
Thank you for following this bug through to resolution, and congratulations on your first contribution to the code base.

Can you please write a few words to explain the change for end users in CHANGELOG.md? Use the prefix: `Fixed: ` in the `[unreleased]` section

I've changed the destination branch of this PR to `develop`, we use `main` for released versions. Please do a pull from `develop` to get the latest commits to avoid causing a merge conflict on `CHANGELOG.md`.

waynepiekarski:
Thanks for the comments, I have rebased on develop and added a line to the CHANGELOG.md

PR #586, opened 2026-05-08
Add facing wizard

Lyrkan:
Another thing I've been working on lately.
Still a WIP but I'd really like to have some feedback about it.

image

image

image

image

image

image

image

image

image

image

image

PR #590, opened 2026-05-09
Gcode viewer: Fit the viewer to the path's bounding box, not its max X/Y/Z

Lyrkan:
Fixes an issue I noticed when working on #586 with a top-right origin operation.

Having a file with small or only negative X/Y values could cause the viewer scaling to be really bad since it was based on `max(max(x), max(y), max(z))`. Relying on the largest span instead should give better results.

For instance, when loading [this file](https://github.com/user-attachments/files/27552081/facing_wizard_top_right.txt) :

**Before**
image

**After**
image


SergeBakharev:
Love your work

PR #591, opened 2026-05-09
File viewer: Fix character sometimes missing in the last line

Lyrkan:
I'm not sure why it was done this way but the current code stripped one character at the end of each line (for `\n`) before removing `\x0d` (= `\r`).

This doesn't work if there is no empty line at the end of the file since it will instead remove the last character of the last line.
Instead we can just strip both `\r` and `\n`.

For instance using [this file](https://github.com/user-attachments/files/27552535/facing_wizard_top_right.txt):

**Before**
image

**After**
image


SergeBakharev:
Nice one!

Issue #157, opened 2025-03-21
Add the ability to set custom tool numbers through the set and change tool dropdowns


faecorrigan:
The stock firmware allows up to tool 256, with plans to expand to 1000
the community firmware allows tools up to 99999

![Image](https://github.com/user-attachments/assets/32b243e6-6e6f-4bae-9b97-e00183f500f9)

![Image](https://github.com/user-attachments/assets/a3b275be-075b-4ac5-8275-59f6a1f09221)

PR #592, opened 2026-05-15
Probe Scan tool

Lyrkan:
This PR adds a new tool providing basic CMM functionality.

It allows users to create features from either direct measurements or derived operations (segments, intersections, midpoints, tangents, circumcircles, open/closed polygons). These can be exported as JSON (to resume sessions later), CSV, or DXF.

Everything currently operates in 2D (X/Y axes) using WCS coordinates. I considered adding support for MCS coordinates since it may be useful to change the WCS for some measurements (such as probing corners at an angle), but I'm not sure if it’s worth the added complexity at that point.

The result of the probe operations is retrieved through an observer pattern implemented in `serial_listeners.py` that allows other modules to register a method that will be called when anything is received from the machine. In this case the probe scan tool will send some `M118.1` commands after probing, asking for the value of the automatically set variables, and then detect the responses later on.

Since space got a little tight I handled jogging by adding a modal on top of the tool (to avoid constant closing and reopening). I'm not entirely convinced by that solution so I'm open to suggestions.

I'll add a more complete description later on but the screenshots should be self-explanatory.

image

image

image

image

image

image

image

image


Lyrkan:
Putting this as ready before I fall into feature creep.

I updated the screenshots in the initial post so they should reflect what is currently implemented.

Some notes about the changes I made since then:

* Regarding the save/load dialogs: I extracted some of the code and Kivy components we had in `main.py` for the local file upload into `Utils.py` and the `ui` directory so I could reuse them. In theory it should have no impact but we'll have to check if everything works fine on Android.
* Still about the save/load features: I chose to disable them on iOS for now... this is annoying since it makes the probe scan tool a lot less interesting but sadly it seems to require some special handling and I don't own any iOS devices
* For the "probe angle" operations: since the firmware only saves the probed angle and not any of the touched points I chose to represent them as a label positioned at the ending coordinates of the probe operation (so, not exactly where the angle was measured). Maybe we could implement some `#157` and `#158` variables later on to represent them by two segments + a dimension instead.
* The "boss" and "bore" operations result will differ depending on which variant is chosen:
  * for "boss" and "bore" (not rectangular with both X and Y axes): this will create an ellipse or a circle (if the delta between the two measured diameters is less than a configurable tolerance)
  * for an X-only or Y-only measurement: this will create an aligned segment with a midpoint
  * for a rectangular probe with both X and Y axes: this will create two segments with a midpoint

Lyrkan:
@SergeBakharev FYI while cleaning up some of my changes I noticed a couple potential issues with the `is_jogging_enabled()` method on `develop`:

https://github.com/Carvera-Community/Carvera_Controller/blob/aae363179ddf6b3c23b3f649a5fbf3cbb8bd1d17/carveracontroller/main.py#L5798-L5807

1) The first check allows jogging if the machine is playing a file (and not paused) which I don't think should be the case
2) The first condition in the final `return` is `not app.playing` which means that `(app.playing and app.state == 'Pause')` on the next line will never be used (since `app.playing` cannot be `true` anymore at that point).
3) If `allow_jogging_while_machine_running` is not enabled it goes into the final `return` which allows jogging if the current state is `Run`

I've fixed this on this branch in [this commit](https://github.com/Carvera-Community/Carvera_Controller/pull/592/commits/8b0b1b962394a5aff89076c79f291438cd9a2e2a), do you want it to be submitted in a separate PR as well (in case this one doesn't make it into the next release)?

SergeBakharev:
Good spot. If it's not too much trouble, cherry-picking would be awesome. Otherwise I'm happy to create an issue to remind us to do it closer to the release cut off

Lyrkan:
> Good spot. If it's not too much trouble, cherry-picking would be awesome. Otherwise I'm happy to create an issue to remind us to do it closer to the release cut off

Done: https://github.com/Carvera-Community/Carvera_Controller/pull/599

Not a cherry-pick since the method on this branch is not exactly the same, but the merge conflict will be easy to resolve.

SergeBakharev:
This is so awesome. I've got a project that needs this too!

The serial observer is probably the best we can do in this bad situation, thank you for contributing this code. Next developer meeting we should have a discussion about taking steps to improve the communication interfaces. It's hard because we have both IP and Serial based comms to support.

I think for a lot of the UI choices we just need to use and iterate as we have more real usage experience. The initial UI looks great. 

Supporting iOS can be put on the backburner since Software-2 was planning to look into improving it.

PR #601, opened 2026-05-22
Show tool change markers on the playback progress bar

Lyrkan:
Just a proposal to add tool change markers to the progress bar.
Marking it as a draft since I still have a few things to check.

image

image


carveracontroller/main.py
```diff
@@ -6490,11 +6511,20 @@ def load_gcode_file(self, filepath):
...
                 self.cnc.parseLine(line, line_no)
                 if self.upcoming_tool == 0:
                     self.upcoming_tool = self.cnc.tool
                 if self.cnc.tool not in self.used_tools:
                     self.used_tools.append(self.cnc.tool)
+                tool_change_label = None
+                if self.cnc.mval == 321:
+                    tool_change_label = 'L'
+                elif self.cnc.tool >= 1 and self.cnc.tool != prev_tool:
+                    tool_change_label = 'T%d' % self.cnc.tool
```
Lyrkan, line 6561: There may be an issue here that I'm not sure how to handle yet: in the CNC parser `M321` sets the value of `tool` to `7` for some reason... so if we go from laser to tool 7 the change will not be detected.

SergeBakharev: I had a quick look at the firmware. 

On the C1 because it has the laser module in the spindle head, M321 sets the value of the current tool to -1 which is the same value for when it's empty.

On the CA1 because the laser is a separate tool that is held in the collet it sets it to tool number 8888.

Not sure why the parser is setting it to 7. Probably some legacy code, and since the 200ms update parser gets the real tool number the value just gets overwritten.

Lyrkan: > since the 200ms update parser gets the real tool number the value just gets overwritten.

I don't think that's it, this index is associated with the toolpaths used by the viewer so it's different from the current tool value retrieved from the machine.

The explanation might be simple: in the Gcode viewer toolbar the laser is the 7th button, so they just used 7 as a number to toggle it on/off.

https://github.com/Carvera-Community/Carvera_Controller/blob/cf2df4a0bb5db385b546db617f16c042e42f3428/carveracontroller/makera.kv#L4271-L4281

This means that it is already a bit broken since having an actual `T7` in the gcode incorrectly shows the laser button (which then controls those paths). For instance here I don't have any `M321` in my file but the laser button is active and controls the first path's visibility:

image

Maybe we could use `8888` instead as a quick fix. It won't match the C1 but probably won't cause any issue unless someone actually uses `T8888`.

Lyrkan: I opened [an issue](https://github.com/Carvera-Community/Carvera_Controller/issues/606) to fix this in another PR.

Lyrkan:
Regarding the last commit: I changed the progress bar so it now uses the reported line number (`playedlines`) instead of the reported percentage (`playedpercent`). This doesn't feel right but the issue is that in the firmware the second value is [byte-based](https://github.com/Carvera-Community/Carvera_Community_Firmware/blob/6cd5ed7be07a5e9a1e1302dd3aef19f8b5909401/src/modules/utils/player/Player.cpp#L642), which could misplace markers since their positions are based on line numbers.

If this is an issue though I can switch back to `playedpercent` but this means that this feature would require us to map line numbers to "% bytes" when preprocessing the file in the controller.

SergeBakharev:
I like this, and agree with the change to using playedlines for the %.

Let's add a Controller setting toggle to show/hide these change markers (can default to show) and let's use this for a bit to see if there are any UX issues.

Lyrkan:
> Let's add a Controller setting toggle to show/hide these change markers (can default to show) and let's use this for a bit to see if there are any UX issues.

Done!

I did a few tests and using the current line it matches the tool markers perfectly.

However I think there is a separate pre-existing issue with `playedlines`: it seems to go backward sometimes. For instance just after a tool change the selected line in the file viewer and the progress bar go back a few lines for some reason.  I'll see if I can find out why it does that in a separate PR.

SergeBakharev:
@acj was seeing that played lines going backwards behaviour. I thought we found the resolved the issue in the firmware

Lyrkan:
> @acj was seeing that played lines going backwards behaviour. I thought we found the resolved the issue in the firmware

Are you thinking about https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/229 ?

It doesn't seem to jump as much as what's visible in the first video anymore, I only saw it happen right after tool changes. 

Writing this makes me think I should check if it also happens with manual tool changes, during my tests I was using the ATC, it could be related.

Lyrkan:
I think I found the reason: https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/316

PR #603, opened 2026-05-24
Use gcode loaded in the controller when resuming playback at line

### LINKED ISSUE #554: OS Temp directory cleanup clears cached gcode file
per #528 It's possible for the OS to come in and clean up the temp directory while the Controller is still running. This breaks the resume-at-line functionality since it needs to re-read the file to do some specific parsing.

There wasn't sufficient time to address this properly in 2.1.0 except to give the user an error popup and to support retaining the resume-at-line numbers when the same file is selected a second time. But a more robust solution should be implemented in a subsequent release. Either:

1. Maintain an open file handle on the cached version of a selected file while it's still selected to prevent external processes from cleaning it up. Requires revising all the file access that might be done by resume-at-line

or 

2. Use the gcode that is already in memory as part of the UI. Maybe storing it in a more robust way to support both use cases.

Lyrkan:
This should fix #554 by using the lines loaded in the controller instead of relying on the temporary cache file.

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -4160,7 +4160,7 @@ class Makera(RelativeLayout):
         filepath = self.file_popup.local_rv.curr_selected_file
         app = App.get_running_app()
         app.selected_local_filename = filepath
-
+        app.selected_remote_filename = ''
 
         self.file_popup.dismiss()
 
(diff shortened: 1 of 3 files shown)


carveracontroller/main.py
```diff
@@ -4160,7 +4160,7 @@ def view_local_file(self):
         filepath = self.file_popup.local_rv.curr_selected_file
         app = App.get_running_app()
         app.selected_local_filename = filepath
-
+        app.selected_remote_filename = ''
```
Lyrkan, line 4163: This change fixes what (I believe) was a preexisting issue that became a bit more annoying here due to the conditions I used in `_resume_gcode_lines_available` (basically if you loaded a file on the machine/controller and then opened another file locally, the resume feature was still available but applied on the wrong file...).

Feel free to let me know if I missed a reason why we shouldn't reset it here!

SergeBakharev: Good spot. I forgot about this code path to opening files.

PR #605, opened 2026-05-25
Gcode viewer: Clean up + some improvements

Lyrkan:
Opening this as a WIP.

My initial goal was to improve a few things in the 3D viewer (see below) but given its state I ended up doing a lot of (heavily LLM-assisted) refactoring work, for instance (but not limited to):

* moving the shaders to their own directory with proper names (no, `hello_cube.glsl`  did not actually draw a cube :slightly_smiling_face:)
* having only one file for each shader instead of one file dedicated to Android + almost the same file for the other platforms (I've yet to test that it works properly, but on paper it should work)
* removing a lot of commented code / useless comments ([some of them](https://github.com/Carvera-Community/Carvera_Controller/blob/313510521bee321e6f5995e6395dda8efff4a26f/carveracontroller/hello_cube.glsl#L69) being quite funny)
* translating some comments from Chinese to English

Regarding the improvements, a video probably speaks a thousand words (the color schemes thing being greatly inspired by what 3D printing slicers usually permit):

Recording 2026-05-25 at 19 14 26


SergeBakharev:
Awesome! This looks great. Maybe in future we could even add the [machine models](https://github.com/Carvera-Community/Carvera_Community_Profiles/releases/tag/v1.4.3fusion) or 3d bed as a toggle to the visualization. Since we infer the MCS to model location values based on the anchor 1 config setting.

Lyrkan:
Just tested it on Android (had to [pin Cython to v3.0.12 and bump P4A to commit `4174efe`](https://github.com/Lyrkan/Carvera_Controller/commit/6ae06c87d0a739a4a8bfc8b8e491b0c92bdc00a1) to fix the Android build, see resulting job [here](https://github.com/Lyrkan/Carvera_Controller/actions/runs/26447517872/job/77867372848)) with a really old Nexus 7 (2014), and after a small fix it seems to be working fine!

image

PR #607, opened 2026-05-26
Refactor Set Axis/Tool popups out of large files

software-2:
This is the first part of starting to rip stuff out of the large monolithic files. The result is hundreds of lines pulled out of main.py and makera.kv!

Please feel free to push back against any choices I've made here - I'd rather reach a consensus on how to do this in a way that makes sense to the folks who have been and will continue to make large feature changes - if this approach doesn't work for you, please say so!

My approach is to do each of these splits in three parts.

First, I added some visual regression tests. Check out this commit, and run the script to get screenshots of the popups being refactored (use ` --update-references`). Then jump to the refactored commits and run again to do the diff. This is my safeguard to help make sure I make zero visual changes during these refactors.

Second is the move out of the monoliths into their relevant areas. UI is kept in the /ui/ folder, things that interface with an actual machine are kept in /machine/, and previous commits added enforcement to help keep these layers separated. As much as possible for this commit, things are kept as much the same as possible, barring things like PEP 8 compliance where renaming are done in the same commit.

And the third section is where I make changes that I think should be made, but aren't strictly in the spirit of splitting out the functionality. In this case, there was a separate popup defined over and over and over and over again where it was just a little text being changed. Duplicate value validation, duplicate Kivy definitions... we can eliminate so much extra code. I haven't written any functionality tests, so there's a risk of breaking things in a refactor like this.


SergeBakharev:
I like the structure!

PR #610, opened 2026-05-27
Throw error if WHB04 is enabled but permissions are not configured correctly

waynepiekarski:
Previously would silently just ignore connected pendant without giving the user any feedback on the failure. Now we give a detailed message to the user on how to correct the problem.

Tested with LHB04B-4 pendant on a Debian 13 x86_64 laptop. Debian does not configure permissions for the pendant, they are root:root by default.

### DIFF
--- a/carveracontroller/addons/pendant/pendant.py
+++ b/carveracontroller/addons/pendant/pendant.py
@@ -140,6 +140,7 @@ if WHB04_SUPPORTED:
             self._daemon.on_jog = self._handle_jogging
             self._daemon.on_button_press = self._handle_button_press
             self._daemon.on_stop_jog = self._handle_stop_jog
+            self._daemon.on_permission_error = self._handle_permission_error
 
             self._daemon.start()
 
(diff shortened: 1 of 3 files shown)


carveracontroller/addons/pendant/pendant.py
```diff
@@ -320,6 +321,49 @@ def _handle_stop_jog(self, daemon: whb04.Daemon) -> None:
...
+                "To fix this on Linux, create the file:\n"
+                "  /etc/udev/rules.d/99-whb04b-pendant.rules\n\n"
+                "containing:\n"
+                f"  {rule}\n\n"
+                "Then run:\n"
+                "  sudo udevadm control --reload-rules\n"
+                "  sudo udevadm trigger\n\n"
+                "Unplug and re-plug the pendant, then restart the application."
+            )
+            print(f"\nERROR: {message}\n", flush=True)
```
Lyrkan, line 331: I think it would be better to redirect the user to [the documentation](https://carvera-community.gitbook.io/docs/controller/features/pendant-support#linux). It would make it easier to update it / fix it if needed later on.

SergeBakharev: I agree with @Lyrkan best to redirect the user to the docs than hardcode instructions.

waynepiekarski: I have updated this to only link to https://carvera-community.gitbook.io/docs/controller/features/pendant-support#linux

waynepiekarski:
Updated to use documentation link, added CHANGELOG.md entry

PR #611, opened 2026-05-27
GCode parser: Do not set tool number to 7 after M321

### LINKED ISSUE #606: GCode parser treats laser as tool 7
Opening this issue following my comment [here](https://github.com/Carvera-Community/Carvera_Controller/pull/601#pullrequestreview-4342272075).

Currently the GCode parser will detect M321 calls and associate tool 7 to the paths that follow.
This tool number doesn't really seem to be used apart for the visibility buttons of the viewer (the laser being the 7th button).

It also doesn't match what we have in the rest of the code (controller and firmware) where the laser is actually T8888 for the CA1 or -1 for the C1.

Since a file can have an actual T7 it can cause some conflicts, for instance in this file where the laser button is enabled and controls the first toolpath:

Image

Maybe we could use 8888 instead as a quick fix (both in the parser and in the toolbar). It won't match the C1 but probably won't cause any issue unless someone actually uses T8888.

Lyrkan:
Fixes #606.

The laser will now be treated as tool 8888. Even though that tool number is used on the CA1 and not on the C1 it should not cause a big issue unless someone actually uses a tool with that same number (which is a lot less likely than what we currently have with tool 7).

I also:
- made it so the probe / 3D probe are shown in the progress bar / color scheme panel if they are used in the currently loaded file (see screenshots)
- moved those special tool numbers to constants (`LASER_TOOL_NUMBER`, `PROBE_TOOL_NUMBER`, `PROBE_3D_TOOL_NUMBER`) instead of hardcoding them here and there
- unified the two color palettes used in the progress bar and color scheme panel. They were added in two different PRs but I actually used similar colors (one being a pastel version of the other)
- Unrelated but something I noticed while testing: the facing tool was not setting `selected_local_filename` when previewing a file. It worked but some things were disabled due to that.

fix-606-laser

fix-606-Probe

fix-606-3DProbe


carveracontroller/main.py
```diff
@@ -4023,11 +4023,11 @@ def open_tool_confirm_popup(self):
...
         target_collet_type = CNC.vars['target_collet_type']
         target_collet_type_text = ["Undefined","3mm", "1/8\"", "4mm", "6mm", "1/4\"","8mm"]
-        if CNC.vars['target_tool'] == 0:
+        if CNC.vars['target_tool'] == PROBE_TOOL_NUMBER:
             target_tool = 'Probe'
-        elif CNC.vars['target_tool'] == 8888:
+        elif CNC.vars['target_tool'] == LASER_TOOL_NUMBER:
             target_tool = 'Laser'
-        elif CNC.vars['target_tool'] >= 999990 and CNC.vars['target_tool'] <= 999999:
+        elif CNC.vars['target_tool'] >= PROBE_3D_TOOL_NUMBER and CNC.vars['target_tool'] <= 999999:
```
Lyrkan: I'm not sure to understand why we have a `CNC.vars['target_tool'] <= 999999` only  there, or why we do `>= 999990`  instead of just `== 999990`, so I didn't change that.

SergeBakharev: The tools with numbers 999990-999999 are treated as "probe-like" in the firmware, and have special handling. 

The reason it's not a single fixed number was the idea that different probes could exist in the system. This hasn't been explored beyond personal hw prototypes, so how it's represented in the UI is not fully mature.

Lyrkan: I added some `PROBE_TOOLS_RANGE_START` and `PROBE_TOOLS_RANGE_END` constants to represent that.

I still use `PROBE_3D_TOOL_NUMBER` when we are actually checking for the 3D probe.

Couldn't really test the change though since my machine just died...

PR #613, opened 2026-06-01
Gcode Viewer: View cube + ortho projection + height-based color scheme + more clean-up

Lyrkan:
Some new improvements for the GCode viewer:

* Added a view cube that shows the current orientation and allows to directly rotate the view to show a given face
* Added a new toggle that allows to switch to an orthographic projection. I've got a slight concern here that the toolbar may start to become a bit full on small screen devices. I'm not sure how we could improve the UI to avoid that. Also no idea what kind of icon I could use for this so the button just says "Ortho" :sweat_smile: 
* Added a height-base color scheme to the previously implemented panel. Note that it doesn't directly take the max/min Z-height for the upper/lower limits since retract/clearance moves would cause all the other moves to become blue most of the time. Instead it excludes outliers (based on percentiles) so the scale actually focus on the actual moves (see video below)
* Did a new clean-up pass since a lot of code (about 700 lines if I recall correctly)  was actually dead, and moved the viewer's assets into their own directory in `carveracontroller/data/`

view-cube-ortho-height


SergeBakharev:
Love your work

PR #621, opened 2026-06-05
Fix unhandled ValueError causing WHB04 pendant feedrate crash Issue #620

### LINKED ISSUE #620: Unhandled exception sometimes when Pendant on and gcode running
I get the following sometimes when my Pendant is on (not using) and gcode is running. Causes the Controller to crash.

unhandled exception: Feedrate outside range
Traceback (most recent call last):
  File "__main__.py", line 12, in <module>
  File "carveracontroller\main.py", line 6471, in main
    global_app.run()
  File "kivy\app.py", line 956, in run
  File "kivy\base.py", line 574, in runTouchApp
  File "kivy\base.py", line 339, in mainloop
  File "kivy\base.py", line 379, in idle
  File "kivy\clock.py", line 733, in tick
  File "kivy\clock.py", line 776, in post_idle
  File "kivy\\_clock.pyx", line 620, in kivy._clock.CyClockBase._process_events
  File "kivy\\_clock.pyx", line 653, in kivy._clock.CyClockBase._process_events
  File "kivy\\_clock.pyx", line 649, in kivy._clock.CyClockBase._process_events
  File "kivy\\_clock.pyx", line 218, in kivy._clock.ClockEvent.tick
  File "carveracontroller\addons\pendant\pendant.py", line 80, in <lambda>
    Clock.schedule_once(lambda _: f(), 0)
                                  ^^^
  File "carveracontroller\addons\pendant\whb04.py", line 377, in <lambda>
    self.callback_executor(lambda: self.on_update(self))
                                   ^^^^^^^^^^^^^^^^^^^^
  File "carveracontroller\addons\pendant\pendant.py", line 146, in _handle_display_update
    daemon.set_display_feedrate(self._cnc.vars["curfeed"])
  File "carveracontroller\addons\pendant\whb04.py", line 280, in set_display_feedrate
    raise ValueError("Feedrate outside range")
ValueError: Feedrate outside range

I was going to to take a look to see if I could fix the unhandled exception.

rbeard-ewa:
The Issue:
When executing dynamic 4th-axis toolpaths (such as continuous rotary finishing operations), the machine's internal curfeed variable can spike significantly as the firmware dynamically calculates angular rotational speed to maintain a constant linear surface speed. Values can easily fluctuate between single digits and 8000+ mid-operation.

When the background Kivy loop (_handle_display_update in pendant.py) polls this curfeed value and passes it to the WHB04 driver, it occasionally exceeds the formatting limits of the physical  display. The driver explicitly raises a ValueError("Feedrate outside range"). Because this exception was unhandled in the clock callback, it would bubble up and completely crash the Kivy main loop, killing the controller UI while the machine was running.

The Fix:

Safely parsed the curfeed variable.
Clamped the raw feedrate to 9999.0 (the maximum displayable value for the 4-digit WHB04 segment).
Wrapped the daemon.set_display_feedrate() call in a try...except ValueError block.
If the driver still rejects the value, the exception is caught silently. This allows the Kivy event loop to continue polling and keeps the UI (and the other functioning coordinate readouts on the pendant) completely stable.

Testing Performed:
Validated locally on a Carvera Air running continuous 4th-axis multi-operation toolpaths via DeskProto.
Confirmed the Kivy UI loop remained perfectly stable during extreme curfeed fluctuations and finished the complete cut without a single crash.

Fixes #620

### DIFF
--- a/carveracontroller/addons/pendant/pendant.py
+++ b/carveracontroller/addons/pendant/pendant.py
@@ -143,7 +143,15 @@ if WHB04_SUPPORTED:
             daemon.set_display_position(whb04.Axis.Y, self._cnc.vars["wy"])
             daemon.set_display_position(whb04.Axis.Z, self._cnc.vars["wz"])
             daemon.set_display_position(whb04.Axis.A, self._cnc.vars["wa"])
-            daemon.set_display_feedrate(self._cnc.vars["curfeed"])
+            # daemon.set_display_feedrate(self._cnc.vars["curfeed"])
+            # rbeard-ewa fix unhandled exception for bad feedrate
+            try:
+              raw_feed = float(self._cnc.vars.get("curfeed", 0))
+              safe_feed = max(0.0, min(raw_feed, 9999.0))
+              daemon.set_display_feedrate(safe_feed)
+            except ValueError:
+              pass
+            # end of feedrate fix
             daemon.set_display_spindle_speed(self._cnc.vars["curspindle"])
             
             # Update the step indicator to reflect current jog mode


SergeBakharev:
Good investigating! This is as a result of removing some artificial 4th axis limits as described: https://carvera-community.gitbook.io/docs/updates/firmware-controller-2.1.0c

Around A (helix) and waterline/contour operations on the 4th axis will benefit the most from the speed ups, which explains why we didn't seem this issue earlier.

SergeBakharev:
Oh whoops, I forgot to check if you had updated the CHANGELOG.md as part of this PR. It's okay, I'll add the item to the file myself.

SergeBakharev:
And I merged it into main, whoops. We collect our commits for the next release in `develop`.

I'll fix this up

carveracontroller/addons/pendant/pendant.py
```diff
@@ -143,7 +143,15 @@ def _handle_display_update(self, daemon: whb04.Daemon) -> None:
...
             daemon.set_display_position(whb04.Axis.A, self._cnc.vars["wa"])
-            daemon.set_display_feedrate(self._cnc.vars["curfeed"])
+            # daemon.set_display_feedrate(self._cnc.vars["curfeed"])
+            # rbeard-ewa fix unhandled exception for bad feedrate
+            try:
+              raw_feed = float(self._cnc.vars.get("curfeed", 0))
+              safe_feed = max(0.0, min(raw_feed, 9999.0))
+              daemon.set_display_feedrate(safe_feed)
+            except ValueError:
+              pass
```
Lyrkan, line 153: Not sure about the `try`/`except`/`pass` here.

If the feed is capped before calling `set_display_feedrate` there is no reason for it to throw a ValueError anymore. If it does then it means something else went wrong and we probably don't want it to just be ignored.

Maybe a good idea though would be to keep the `try`/`except` but:
- put it around all the `set_display_*` calls (since they can all throw a `ValueError`
- add a logger call instead of the `pass` so we can at least see that there was an issue without the app crashing due to the handled exception

rbeard-ewa: Good points, Vincent. I mostly just wanted to get this specific crash out of the way so I could get back to using the machine. Since I don't have the bandwidth to thoroughly test the other set_display calls, I'll leave it to you or the core team to implement the logging and broader exception handling if you'd like. Feel free to modify this or open a new PR on top of it!

PR #626, opened 2026-06-08
Remote file multiselect

danilius:
Adds desktop-style multi-select support to the remote file list in the Select File dialog.

Changes:
- Normal click selects one remote item
- Ctrl-click toggles individual remote items
- Shift-click selects a range
- Delete supports multiple selected files/directories
- Rename and Select remain single-item actions
- Multi-delete sends remote delete commands sequentially

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 [unreleased]
-- Enhancement: Add multi-select deletion to the remote file browser
+- Enhancement: Add multi-select to the remote file browser
 - Enhancement: Display error message in halt popup. Requires halt errors to start with "ERROR: " in the firmware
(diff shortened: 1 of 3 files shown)


SergeBakharev:
Can you please fill in a line in CHANGELOG.md that described the change? Put it in the `[unreleased]` section, thats where we keep the new changes not yet in a stable release. You can prefix the line with `Enhancement: `

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 [unreleased]
+- Enhancement: Add multi-select deletion to the remote file browser
```
SergeBakharev: It's not just deletion though

```suggestion
- Enhancement: Add multi-select to the remote file browser
```

Issue #637, opened 2026-06-27
Rotation value not showing up in WCS tab


rbeard-ewa:
The rotation value was not showing up for me under the WCS value like it says in the documentation. Maybe this is now intentional as it seems as if it was removed and the documentation shows that it used to be there.

I put this in my copy as I would like to see the value. Here is a snippet from updateStatus in main.py:

            rotation_angle = CNC.vars["rotation_angle"]
            desc_key = coord_system_name.lower().replace(".", "_") + "_description"
            try:
                wcs_description = Config.get("carvera", desc_key).strip()
            except Exception:
                wcs_description = ""
            if wcs_description:
                self.coord_system_data_view.main_text = wcs_description
            else:
                self.coord_system_data_view.main_text = coord_system_name
            # ewabob put out angle when WCS rotated    
            if abs(rotation_angle) > 0.001:
                self.coord_system_data_view.minr_text = f"R: {rotation_angle:.3f}°"
            else:
                self.coord_system_data_view.minr_text = coord_system_name
           # end of ewabob**    

Also stumbled across this when looking into why I didn't see an angle being output. Maybe this is ok but looks wrong to me:

 def update_ui(dt):
            # wcs_data format: {'G54': [x, y, z, a, rotation], 'G55': [...], ...}
            
            for wcs, values in wcs_data.items():
                if len(values) >= 5:  # Ensure we have X, Y, Z, A, rotation
                    x, y, z, a, b, rotation = values

Code checks for 5 or greater values and then the assignment demands 6 values from the uppacking. I didn't see b being used below the code.

SergeBakharev:
Can you please open a Pull Request with your code changes? Please target th Dev branch as your destination.

> I didn't see b being used below the code.

That is correct, while the firmware technically supports 5 axis, the Controller does not.

rbeard-ewa:
Lyrkan says this is intentional.

SergeBakharev:
Continuing the discussion from #638 

Good observations @Lyrkan  and @rbeard-ewa . We are overloading the value being shown in the minor text of the button for two purposes, and this is in conflict with our documentation.

I think the right approach is to follow the pattern of the other buttons, and take turns showing both values.

PR #638, opened 2026-06-27
Fix wcs rotation display

rbeard-ewa:
Fix missing WCS rotation angle display in UI. Currently if there is a rotation angle in the WCS, it just outputs coord_system_name instead of the actual angle.

Issue #637

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -5077,7 +5077,14 @@ class Makera(RelativeLayout):
                 self.coord_system_data_view.main_text = wcs_description
             else:
                 self.coord_system_data_view.main_text = coord_system_name
-            self.coord_system_data_view.minr_text = coord_system_name
+                
+            # ewabob put out angle when WCS rotated    
+            if abs(rotation_angle) > 0.001:
+                self.coord_system_data_view.minr_text = f"R: {rotation_angle:.3f}°"
+            else:
+                self.coord_system_data_view.minr_text = coord_system_name
+           # end of ewabob    
+            
             self.coord_system_data_view.scale = 80.0 if abs(rotation_angle) > 0.01 else 100.0
             
             # Update WCS Settings popup if it's open


carveracontroller/main.py
```diff
@@ -5769,61 +5077,68 @@ def updateStatus(self, *args):
...
             else:
                 self.coord_system_data_view.main_text = coord_system_name
-            self.coord_system_data_view.minr_text = coord_system_name
+                
+            # ewabob put out angle when WCS rotated    
+            if abs(rotation_angle) > 0.001:
+                self.coord_system_data_view.minr_text = f"R: {rotation_angle:.3f}°"
+            else:
+                self.coord_system_data_view.minr_text = coord_system_name
+           # end of ewabob
```
Lyrkan, line 5086: I'd suggest removing this comment. If we added similar ones every time something changed, the codebase would get cluttered pretty quickly.

Same goes for adding your name at the beginning of the first one, it is not really necessary (we can find out who changed a line using git blame) and could even end-up being a bit confusing later on since it doesn't match your Github handle.

Lyrkan (commented):
Looking at https://github.com/Carvera-Community/Carvera_Controller/pull/511 it seemed intentional.

If we merge this it means that when a description is set the current WCS name won't be displayed on the main screen anymore, which could end up being a bit annoying in some cases.

If we really wanted to display the current WCS rotation I think it should at least be in addition to the name by cycling between the two values (like it's done in some of the other buttons).

rbeard-ewa:
Closing request because Lyrkan says the WCS name will disappear. Doesn't for me but he must have tried it and it does for him.

Lyrkan:
@rbeard-ewa No need to close this pull request, especially if you think I may be wrong.

My assumptions were based on reading the code, and unless I missed something it would work this way:
* Main text: WCS description (`wcs_description`) if it is set, else fallback to WCS name (`coord_system_name`)
* Minor text: WCS rotation (`rotation_angle`) if it is set, else fallback to WCS name (`coord_system_name`)

Let's say you were using G54, you'd get:

| `wcs_description`                 | `rotation_angle` | Main text                     | Minor text |
|-------------------------------|-----------------|-------------------------------|------------|
| `<not set>`                      | `<not set>` | G54                           | G54        |
| Custom description of the WCS | `<not set>` | Custom description of the WCS | G54        |
| `<not set>`                   | 1.23°       | G54                           | 1.23°      |
| Custom description of the WCS | 1.23°       | Custom description of the WCS | 1.23°      |

As you can see it'd mean that `G54` wouldn't be displayed at all for the last row.

On the other hand people could put the name in the description in this case, so maybe it isn't as big of an issue, but that's still something that should be discussed in my opinion.

PR #674, opened 2026-07-13
Fix resume-at-line WCS recovery after G53

### LINKED ISSUE #665: Resume-at-line treats non-modal G53 as the active work coordinate system
## Summary

Resume-at-line currently considers `G53` alongside `G54` through `G59.3` when it reconstructs the active work coordinate system. It then replays whichever command appeared most recently.

`G53` is a non-modal qualifier for a machine-coordinate move, not a persistent work-coordinate-system selection. Replaying a bare `G53` therefore does not restore the interrupted program's active WCS.

The behavior is present on `develop` at `2ea66c10e96d88013a04441134d3d4f8e8c88510`: [WCS recovery logic](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/Controller.py#L1122-L1143). LinuxCNC documents `G53` as non-modal and `G54` through `G59.3` as WCS selectors: [coordinate-system documentation](https://linuxcnc.org/docs/stable/html/gcode/coordinates.html). The Community Firmware likewise consumes `G53` as a machine-coordinate flag for the associated movement: [firmware implementation](https://github.com/Carvera-Community/Carvera_Community_Firmware/blob/feab653def96a959e695049a4f828aeddbcd5547/src/modules/communication/GcodeDispatch.cpp#L193-L222).

## Minimal reproduction

```gcode
G55
G53 G0 Z-2
G1 X10 F100
G1 X20
```

Preview a resume at line 4.

Expected:

- The recovery command sequence contains `buffer G55`.
- The controller's intentional safe-Z command remains `buffer G53 G0 Z-2`.

Actual:

- The sequence contains a bare `buffer G53` instead of restoring `G55`.
- The separate safe-Z command is also present.

## Impact

If the machine still retains `G55`, this may not change the resulting position. If the retained WCS differs after a reset, reconnect, or manual WCS change, however, resume positioning can occur in the wrong coordinate frame. Resume should reconstruct the file's modal WCS deterministically rather than depend on retained machine state.

## Acceptance criteria

- Only `G54` through `G59.3` participate in modal WCS recovery.
- The intentional safe-Z move still uses `G53 G0 Z-2`.
- A regression test proves that a later `G53` move does not replace an earlier active WCS.

righteousgambit:
Closes #665

## Summary

Exclude non-modal `G53` from resume-at-line's active-WCS search while preserving the controller's explicit `G53 G0 Z-2` recovery move.

## Changes

- Remove `G53` from the modal WCS candidates used by `playStartLineCommand()`.
- Add a regression test with `G55` followed by a `G53` machine-coordinate move.

## Why

`G53` applies machine coordinates to its associated movement; it does not select a persistent coordinate system. Treating it as the most recent WCS can prevent resume-at-line from restoring the program's actual `G54`-`G59.3` state.

This is state-dependent: if the machine already retains the intended WCS, behavior may be unchanged. The fix removes that dependency and makes resume reconstruction deterministic.

## Validation

- Regression against the pinned `develop` implementation: `1 failed` because the preview contained bare `buffer G53` and omitted the expected `buffer G55`.
- With this change, using the same exact-ref sparse harness: `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=/tmp/carvera-test-stubs:. pytest -q tests/unit/test_resume_wcs.py` -> `1 passed`.

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`d989beb58eaa`](https://github.com/righteousgambit/Carvera_Controller/commit/d989beb58eaafc7792eead19cff4a89fabb8f3a6), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run pytest -q tests/unit/test_resume_wcs.py
```

Result: **1 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..d989beb58eaafc7792eead19cff4a89fabb8f3a6`: **passed**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Enhancement: Advanced TLO Calibration option. Here you can set the offset to use from tool setter, and/or the number of repeat probings to use
 - Enhancement: Added a warning popup if the controller version is lower than the firmware. This is not a supported config
 - Enhancement: Added Auto Ext. Out toggle to spindle dropdown and Config and Run screen. This can be used to automatically run a vacuum or compressor when the spindle is running
+- Fixed: Resume-at-line no longer treats the non-modal G53 command as the active work coordinate system
 - Fixed: Restore Keyboard Jogging state after Probing Popup is closed
 - Fixed: Repeated firmware checks now happen just once
 - Fixed: UI widget updates from the SerialMonitor() now dispatched via the main thread. This should reduce the number of RecycleView related crashes
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -1121,7 +1121,6 @@ class Controller:
 
             # Search for WCS coordinate space - find the last one used
             wcs_commands = [
-                ("G53", self._find_command_line_number(lines, start_line, "G53")),
                 ("G54", self._find_command_line_number(lines, start_line, "G54")),
                 ("G55", self._find_command_line_number(lines, start_line, "G55")),
                 ("G56", self._find_command_line_number(lines, start_line, "G56")),
--- a/tests/unit/test_resume_wcs.py
+++ b/tests/unit/test_resume_wcs.py
@@ -1,0 +1,20 @@
+"""Tests for resume-at-line work coordinate system recovery."""
+
+from carveracontroller.Controller import Controller
+
+
+def test_g53_does_not_replace_active_work_coordinate_system():
+    controller = Controller.__new__(Controller)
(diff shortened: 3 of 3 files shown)


SergeBakharev:
G53 is indeed non-modal. Good spot. PR looks good to me. Merging

PR #675, opened 2026-07-13
Verify XMODEM downloads against advertised MD5

### LINKED ISSUE #666: XMODEM download accepts EOT without verifying the advertised MD5
## Environment

- Branch: `develop`
- Validated revision: [`2ea66c10e96d88013a04441134d3d4f8e8c88510`](https://github.com/Carvera-Community/Carvera_Controller/commit/2ea66c10e96d88013a04441134d3d4f8e8c88510)
- Transfer mode: XMODEM-8K with CRC
- Hardware/firmware: deterministic transport double; no machine or firmware required

## Summary

The receiver reads the machine-provided whole-file MD5 from block zero, but only uses it to decide whether an existing local file can be reused. It does not retain that digest and verify the bytes received during the current transfer. A peer that sends EOT after only a valid prefix therefore gets an ACK and the receiver reports a nonnegative byte count, leaving the caller to treat an incomplete artifact as a successful download.

## Root cause

- [Block zero parses the remote MD5 and only compares it with the optional local digest](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/XMODEM.py#L811-L830).
- [The EOT path ACKs and returns the accumulated padded-byte counter](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/XMODEM.py#L731-L739) without comparing the received content with the advertised digest.
- [`download_file()` treats every nonnegative result as success and deletes the destination only when the receiver returns `None`](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/main.py#L4702-L4718).

## Reproduction

Use a deterministic XMODEM stream containing:

1. Block zero advertising the MD5 of the intended complete file.
2. Block one containing only a truncated prefix.
3. EOT.

On the pinned revision, the receiver ACKs EOT and returns `8192` rather than rejecting the artifact. The regression assertion fails as:

```text
E       assert 8192 is None
```

The same pinned revision also fails to preserve the cache shortcut when the equivalent local MD5 is supplied in uppercase; it proceeds with the transfer instead of returning `0`, leaving the output empty, and issuing three CAN writes.

## Impact

Ordinary TCP packet loss does not manufacture an EOT, so this is specifically an integrity failure when the peer ends the protocol prematurely. The remote `play` command is sent before the download in [`Controller.py`](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/Controller.py#L628-L633), so this reproduction does not establish remote mid-cut execution. It does establish that incomplete local downloads can be accepted and then affect supported local behavior such as previewing, line counting, resuming, or reusing the artifact.

## Acceptance criteria

- Retain the expected MD5 supplied in block zero.
- Incrementally hash only the actual, unpadded payload bytes written to the destination.
- ACK EOT for protocol compatibility, but return `None` when the expected digest is missing or does not match.
- Preserve the block-zero cache shortcut, including case-insensitive comparison of equivalent hexadecimal MD5 strings: return `0`, write no output, and send exactly three CAN bytes.
- Add deterministic regression coverage for matching content, mismatched/truncated content, and the uppercase-local-MD5 cache shortcut.

righteousgambit:
Closes #666

## What changed

- Retain the whole-file MD5 advertised by XMODEM block zero.
- Hash each actual, unpadded payload slice as it is written.
- At EOT, keep the existing ACK behavior but reject the local artifact with `None` when the advertised digest is absent or does not match.
- Normalize hexadecimal digest comparisons so an uppercase local MD5 still takes the existing cache shortcut.
- Add deterministic tests for a matching transfer, a prematurely terminated/truncated transfer, and the uppercase-local-MD5 shortcut (`0`, empty output, three CAN writes).
- Document the integrity fix in the changelog.

## Protocol compatibility

The receiver continues to ACK EOT on the wire. The behavioral change is local: a transfer whose bytes do not match the machine-advertised MD5 is no longer reported as successful to `download_file()`.

## Validation

Pinned `develop` (`2ea66c10e96d88013a04441134d3d4f8e8c88510`) against the new tests:

```text
2 failed, 1 passed
```

The truncated-transfer assertion observes `8192` instead of `None`; the uppercase-local-MD5 test observes a completed transfer instead of the cache short-circuit.

Corrected implementation:

```text
3 passed
```

Quality checks:

```text
ruff check carveracontroller/XMODEM.py tests/unit/test_xmodem_download_integrity.py
ruff format --check carveracontroller/XMODEM.py tests/unit/test_xmodem_download_integrity.py
python -m compileall -q carveracontroller/XMODEM.py tests/unit/test_xmodem_download_integrity.py
git diff --check HEAD^ HEAD
```

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`f96ceccf1615`](https://github.com/righteousgambit/Carvera_Controller/commit/f96ceccf161516a5e02c2eb995b1b7661aba644b), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run python -m pytest -q tests/unit/test_xmodem_download_integrity.py
```

Result: **3 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..f96ceccf161516a5e02c2eb995b1b7661aba644b`: **passed**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed: Reject incomplete downloads that do not match the machine-provided MD5 checksum
 - Enhancement: Add multi-select to the remote file browser
 - Enhancement: Display error message in halt popup. Requires halt errors to start with "ERROR: " in the firmware
 - Enhancement: Add popup notice when using stock firmware instead of the Community Firmware
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
 - Fixed: Ignore unknown WHB04 button values without reconnecting or dropping valid paired inputs
 - Fixed: Resume-at-line restores feed rates from standalone and tightly packed F words before recovery moves
 - Fixed: Resume-at-line no longer treats the non-modal G53 command as the active work coordinate system
-- Fixed: Reject incomplete downloads that do not match the machine-provided MD5 checksum
+- Fixed: Reject downloads whose content does not match the machine-provided MD5. Skip MD5 check when none is available, and defer .lz checks until after decompress
 - Change: Misleading "Download canceled by Controller!" MDI message is suppressed, in logs a message is recorded that cached version of the config.txt was used
 - Change: Remove remaining "Can not load config, Key:" messages from the MDI
 - Change: Resume playback will now use gcode loaded in the controller instead of cached local file
--- a/carveracontroller/USBStream.py
+++ b/carveracontroller/USBStream.py
@@ -1,5 +1,4 @@
 import logging
-import sys
 import time
 
 import serial
@@ -20,11 +19,8 @@ class USBStream:
     # ----------------------------------------------------------------------
     def __init__(self, log_sent_receive=False):
         self.modem = XMODEM(self.getc, self.putc, "xmodem")
-        handler = logging.StreamHandler(sys.stdout)
-        handler.setLevel(logging.WARNING)
-        formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
-        handler.setFormatter(formatter)
-        self.modem.log.addHandler(handler)
+        # Rely on the app/Kivy root logger; do not attach extra StreamHandlers to
+        # the shared "xmodem.XMODEM" logger (USB+WiFi would duplicate every line).
         self.log_sent_receive = log_sent_receive
         self._send_log_buffer = b""
         self._recv_log_buffer = b""
--- a/carveracontroller/WIFIStream.py
+++ b/carveracontroller/WIFIStream.py
@@ -82,12 +82,8 @@ class WIFIStream:
     # ----------------------------------------------------------------------
     def __init__(self, log_sent_receive=False):
         self.modem = XMODEM(self.getc, self.putc, "xmodem8k")
-
-        handler = logging.StreamHandler(sys.stdout)
-        handler.setLevel(logging.WARNING)
-        formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
-        handler.setFormatter(formatter)
-        self.modem.log.addHandler(handler)
+        # Rely on the app/Kivy root logger; do not attach extra StreamHandlers to
+        # the shared "xmodem.XMODEM" logger (USB+WiFi would duplicate every line).
         self.log_sent_receive = log_sent_receive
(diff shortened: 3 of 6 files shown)


SergeBakharev:
After the comms uplift https://github.com/Carvera-Community/Carvera_Controller/pull/686 the PR needs some revision to also be applied to the newer Makera framed protocol.

SergeBakharev:
E-gawd this is a rabbit hole. 

At first glance this PR seems to resolve a real defect, however I'm putting this PR on ice because it looks like Makera intentionally don't check the Md5 hash. The md5 handling in the firmware is broken/mismatching. I've opened an issue here: https://github.com/Carvera-Community/Carvera_Community_Firmware/issues/357

We need to resolve the jank in the firmware before we can do any enforcement in the Controller. We then need to decide what to do in the cross-compatibility scenario when connecting to a machine running Makera firmware. Probably just not do MD5 checking...

SergeBakharev:
Maybe I'm throwing the "baby out with the bathwater" here and we should just have multiple cases for different handling:

1. If the machine returns the MD5 as an empty string we skip the MD5 check. In the Community firmware, we should revise the code to calculating a fresh MD5, but that can come next release.
2. If file received is .lz compressed, we defer the MD5 check until after we have decompressed the file
3. If the file received is not a .lz file we compare it's MD5

With this case handling we would maintain compatibility with Makera firmware.

I'm too tired to implement anything, so I'll sleep on it for now. I'm still rather uncomfortable with the MD5 hash not being calculated live for the file transferred over the wire, but on a device with limited processing power and memory I guess that makes sense.

SergeBakharev:
I've implemented the md5 checking case as described. In the process of double checking, I added a on-machine integrity check feature: https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/358

Amusingly it found that the factory provided files on my system AudreyHepburn.nc and PirateShip.nc don't match their md5 hashes... I'm not adding any special handling for that

The error message now looks like this, with the actual/expected hashes:
image

PR #676, opened 2026-07-13
Send complete XMODEM frames over Wi-Fi

### LINKED ISSUE #667: Wi-Fi XMODEM putc can truncate a frame on a short socket send
## Environment

- Branch: `develop`
- Validated revision: [`2ea66c10e96d88013a04441134d3d4f8e8c88510`](https://github.com/Carvera-Community/Carvera_Controller/commit/2ea66c10e96d88013a04441134d3d4f8e8c88510)
- Transfer mode: XMODEM-8K with CRC over TCP
- Hardware/firmware: deterministic socket double; no machine or firmware required

## Summary

`WIFIStream.putc()` makes one `socket.send()` call for an entire XMODEM frame and returns immediately. Python documents that [`socket.send()` returns the number of bytes sent and that applications are responsible for transmitting any remaining bytes](https://docs.python.org/3.11/library/socket.html#socket.socket.send). A successful call may therefore accept only a prefix of the frame.

The XMODEM sender supplies the complete frame to `putc()` once, ignores the callback's return count, and waits for the receiver's ACK. On a partial TCP write, the rest of that frame is never queued by the controller.

## Root cause

- [`WIFIStream.putc()` returns the result of a single `socket.send(data)` call](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/WIFIStream.py#L160-L161).
- [The XMODEM sender passes the assembled header, payload, and checksum to `putc()` and does not inspect its return value](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/XMODEM.py#L538-L549).
- [It then waits for an ACK and eventually retries or aborts](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/XMODEM.py#L550-L572).
- [`WIFIStream.upload()` uses this callback for XMODEM uploads](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/WIFIStream.py#L163-L168).

## Deterministic reproduction

Use a socket double whose `send()` accepts only 2,048 bytes from a valid 8,199-byte XMODEM-8K CRC frame. Call `WIFIStream.putc(frame)` once and assert that the complete frame reached the simulated wire.

Command run against pinned `develop`:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 \
  PYTHONPATH=/path/to/pinned-develop \
  python3.11 -m pytest -c /dev/null -p no:cacheprovider \
  --rootdir=/tmp --import-mode=importlib tests/unit/test_wifi_stream.py -q
```

Observed failure:

```text
F                                                                        [100%]
E       AssertionError: assert 2048 == 8199
1 failed
```

The socket accepted exactly the configured 2,048-byte prefix, while the complete frame length was 8,199 bytes.

## Impact

This can cause uploads to retry, time out, or fail when the OS accepts only part of a frame in a `send()` call. The XMODEM CRC and ACK/NAK exchange prevent the truncated frame from being silently accepted as valid, so this is a transfer reliability/availability issue rather than demonstrated silent data corruption.

## Expected behavior

`WIFIStream.putc()` should not return successfully until every byte in the supplied XMODEM frame has been sent, or the socket raises an error. Python's [`socket.sendall()`](https://docs.python.org/3.11/library/socket.html#socket.socket.sendall) provides that contract.

## Acceptance criteria

- Send the entire byte string supplied to `WIFIStream.putc()` before reporting success.
- Return the input length so the callback preserves its existing integer success contract.
- Propagate socket errors rather than reporting a complete write.
- Add a deterministic regression using a valid 8,199-byte XMODEM-8K CRC frame and a socket double that models a 2,048-byte short write.

righteousgambit:
Closes #667

## What changed

- Replace the one-shot `socket.send()` in `WIFIStream.putc()` with `socket.sendall()`.
- Return `len(data)` after `sendall()` completes, preserving the XMODEM callback's integer success contract.
- Add a deterministic regression built from a valid 8,199-byte XMODEM-8K CRC frame.
- Document the Wi-Fi transfer reliability fix in the changelog.

## Why `sendall()`

Python documents that [`socket.send()`](https://docs.python.org/3.11/library/socket.html#socket.socket.send) may send only part of the supplied data and leaves the remainder to the application. [`socket.sendall()`](https://docs.python.org/3.11/library/socket.html#socket.socket.sendall) continues until all bytes have been sent or an error occurs, which is the contract `XMODEM.send()` expects from this callback.

## Impact and compatibility

The wire format and XMODEM CRC behavior are unchanged. This closes a reliability gap that could cause retries, timeouts, or transfer failure after a partial socket write. CRC validation prevents a truncated frame from being silently accepted, so the change does not claim to fix demonstrated undetected corruption.

## Validation

Pinned `develop` (`2ea66c10e96d88013a04441134d3d4f8e8c88510`) against the new regression:

```text
F                                                                        [100%]
E       AssertionError: assert 2048 == 8199
1 failed
```

Corrected implementation:

```text
.                                                                        [100%]
1 passed
```

The passing test verifies all of the following:

- the simulated wire receives all 8,199 bytes in order;
- `send()` is never called;
- `sendall()` is called exactly once; and
- `putc()` returns `8199`.

Quality checks:

```text
ruff check carveracontroller/WIFIStream.py tests/unit/test_wifi_stream.py
ruff format --check carveracontroller/WIFIStream.py tests/unit/test_wifi_stream.py
python -m compileall -q carveracontroller/WIFIStream.py tests/unit/test_wifi_stream.py
git diff --check HEAD^ HEAD
```

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`8c9bb6c8ba77`](https://github.com/righteousgambit/Carvera_Controller/commit/8c9bb6c8ba771e47e1101e23b2864001d15ea20d), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run python -m pytest -q tests/unit/test_wifi_stream.py
```

Result: **1 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..8c9bb6c8ba771e47e1101e23b2864001d15ea20d`: **passed**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed: Ensure complete XMODEM packets are written over Wi-Fi
 - Enhancement: Add multi-select to the remote file browser
 - Enhancement: Display error message in halt popup. Requires halt errors to start with "ERROR: " in the firmware
 - Enhancement: Add popup notice when using stock firmware instead of the Community Firmware
--- a/carveracontroller/WIFIStream.py
+++ b/carveracontroller/WIFIStream.py
@@ -158,7 +158,8 @@ class WIFIStream:
         return None
 
     def putc(self, data, timeout=0.5):
-        return self.socket.send(data) or None
+        self.socket.sendall(data)
+        return len(data)
 
     def upload(self, filename, local_md5, callback):
         # do upload
--- a/tests/unit/test_wifi_stream.py
+++ b/tests/unit/test_wifi_stream.py
@@ -1,5 +1,8 @@
 """Tests for WIFIStream socket write semantics."""
 
+import struct
+
+from carveracontroller.protocols.framing import PTYPE_FILE_DATA, build_frame
 from carveracontroller.WIFIStream import WIFIStream
 from carveracontroller.XMODEM import XMODEM
 
@@ -31,11 +34,16 @@ def _xmodem8k_frame():
     return bytes(modem._make_send_header(8192, 1) + framed_payload + modem._make_send_checksum(1, framed_payload))
 
 
-def test_putc_sends_the_complete_xmodem_frame_and_returns_its_length():
+def _makera_file_data_frame():
+    """Full Makera FILE_DATA frame as emitted by XMODEM.send() via putc."""
+    seq = struct.pack(">I", 1)
+    file_data = b"x" * 8192
+    return build_frame(PTYPE_FILE_DATA, seq + file_data)
+
+
+def _assert_putc_writes_complete_frame(frame):
     stream = WIFIStream.__new__(WIFIStream)
     stream.socket = DeterministicShortWriteSocket(short_write_size=2048)
-    frame = _xmodem8k_frame()
-    assert len(frame) == 8199, "fixture must model a complete xmodem8k frame"
 
     result = stream.putc(frame)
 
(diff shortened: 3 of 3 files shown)


SergeBakharev:
Confirming that the issue reported is real, though I wasn't able to find a real-life way of causing it on my machine.

I've revised the code to handle the new makera framed protocol.

Merging

PR #677, opened 2026-07-13
Ignore unknown WHB04 button codes

### LINKED ISSUE #668: WHB04 reconnects and drops paired STOP/RESET input on unknown button codes
## Controller version

`develop` at `2ea66c10e96d88013a04441134d3d4f8e8c88510`. This is controller-side pendant handling; firmware is not involved in the failure.

## Reproduction

Feed the WHB04 parser one valid eight-byte packet containing a valid safety/action button in either button slot and an unknown nonzero value in the other slot:

```python
packet = struct.pack(
    "BBBBBBbB",
    0x04,
    0,
    Button.STOP.value,  # RESET reproduces it too
    0xFF,               # unknown/transient value
    StepSize.LEAD.value,
    Axis.OFF.value,
    0,
    0,
)
daemon._process_input_packet(packet)
```

The same result occurs with the two button slots reversed.

## Expected

Ignore the unknown button value, preserve the valid button from the same packet, and dispatch its callback without cycling the pendant connection.

## Actual

Both nonzero bytes are converted directly to `Button` enum members. `Button(0xFF)` raises `ValueError` before callbacks are dispatched, so the valid STOP/RESET value in the same packet is dropped. In the live device loop that exception reaches the generic retry handler, closes the current HID guard, reports a disconnect, and reconnects.

The parser already tolerates transient unknown rotary-axis and step-selector values, but not unknown button values: [button parsing](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/pendant/whb04.py#L401-L417), [device-loop propagation](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/pendant/whb04.py#L371-L375), [retry/reconnect handler](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/pendant/whb04.py#L330-L352).

## Impact

One malformed, noisy, or transient button byte can produce a visible pendant disconnect/reconnect and can discard a valid STOP or RESET indication paired in that report. This report demonstrates the controller failure deterministically; it does not claim a particular hardware frequency.

## Regression proof

A four-case test matrix covers STOP and RESET with the unknown value in each slot. Against the pinned base it fails all four cases with `ValueError: 255 is not a valid Button`.

righteousgambit:
Closes #668

## Summary

- parse the two WHB04 button slots independently
- ignore unknown nonzero button values, matching the existing defensive behavior for transient axis and step-selector values
- preserve and dispatch any valid button carried by the same packet
- add focused STOP/RESET coverage for both button-slot orderings
- add the unreleased changelog entry

The change is intentionally limited to enum values the controller does not recognize. Known buttons and normal release handling are unchanged.

## Validation

- regression test on the base: **4 failed**, all at `Button(0xFF)`
- regression test with this change: **4 passed**
- `ruff check` passed for the source and test
- `git diff --check` passed

Targeted command:

```shell
poetry run python -m pytest -q tests/unit/test_whb04_input.py
```

## Changelog

`- Fixed: Ignore unknown WHB04 button values without reconnecting or dropping valid paired inputs`

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`51de21689dcf`](https://github.com/righteousgambit/Carvera_Controller/commit/51de21689dcf7bf98c0f2db0a9136a2ada91114e), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run python -m pytest -q tests/unit/test_whb04_input.py
```

Result: **4 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..51de21689dcf7bf98c0f2db0a9136a2ada91114e`: **passed**

### DIFF
--- a/.github/workflows/quality.yaml
+++ b/.github/workflows/quality.yaml
@@ -28,16 +28,17 @@ jobs:
           path: .venv
           key: ${{ runner.os }}-quality-py3.12-${{ hashFiles('poetry.lock') }}
 
-      - name: Install Linux test prerequisites
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y xvfb libgl1 libgles2 libmtdev1
-
       - name: Install poetry and dependencies
         uses: ./.github/actions/bootstrap-poetry
         with:
           os: linux
           install-args: --only main,lint,test
+
+      - name: Install Linux prerequisites
+        run: poetry run ./scripts/install_linux_prereqs.sh
+
+      - name: Install Linux test prerequisites
+        run: sudo apt-get install -y xvfb libgl1 libgles2 libmtdev1
 
       - name: Run quality hooks
         run: poetry run pre-commit run --all-files
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed: Ignore unknown WHB04 button values without reconnecting or dropping valid paired inputs
 - Enhancement: Add multi-select to the remote file browser
 - Enhancement: Display error message in halt popup. Requires halt errors to start with "ERROR: " in the firmware
 - Enhancement: Add popup notice when using stock firmware instead of the Community Firmware
--- a/carveracontroller/addons/pendant/whb04.py
+++ b/carveracontroller/addons/pendant/whb04.py
@@ -408,10 +408,16 @@ class Daemon:
             return
 
         pressed_buttons = set()
-        if button1 != 0:
-            pressed_buttons.add(Button(button1))
-        if button2 != 0:
-            pressed_buttons.add(Button(button2))
+        for raw_button in (button1, button2):
+            if raw_button == 0:
+                continue
+            try:
+                pressed_buttons.add(Button(raw_button))
+            except ValueError:
+                # Fast or noisy inputs can produce transient button values,
+                # just as they do for the rotary controls below. Preserve any
+                # valid button from the same packet instead of reconnecting.
+                pass
         newly_pressed = pressed_buttons - self._pressed_buttons
         newly_released = self._pressed_buttons - pressed_buttons
         self._pressed_buttons = pressed_buttons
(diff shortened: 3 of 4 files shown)


SergeBakharev:
This will fail CI driven tests until we install the appropriate libusb library into the runner. On the Linux runner it needs libhidapi-hidraw0. This would be met by running `scripts/install_linux_prereqs.sh`

Can you please revise the CI workflow for the quality gate to do this?

righteousgambit:
Revised in commit [`dba9568`](https://github.com/righteousgambit/Carvera_Controller/commit/dba9568690d690700acd0bf9afabf7d6576563a6): the Quality Gate now runs `poetry run ./scripts/install_linux_prereqs.sh` after bootstrapping Poetry, while retaining the explicit Xvfb/OpenGL packages needed by the test job.

Local verification:
- `tests/unit/test_whb04_input.py`: **4 passed**
- Ruff check and format: passed
- strict machine-layer mypy: passed
- import-linter: passed
- workflow validation and shell syntax: passed

The exact-head Quality Gate run is currently awaiting upstream approval (`action_required`, zero jobs): https://github.com/Carvera-Community/Carvera_Controller/actions/runs/29295919067

Once approved, I will verify that collection proceeds past the prior native `hidapi` import failure.

SergeBakharev:
I'm not able to produce a real life scenario with my hardware that causes this problem. But maybe if there is radio interference or different model HB pendant the described issue could occur. Certainly the code allows for it.

Not sure what was going with that one unit test's imports, I've simplified it.

Merging

PR #678, opened 2026-07-13
Fix modal feed recovery for resume-at-line

### LINKED ISSUE #669: Resume-at-line can miss the active feed rate before the recovery Z move
## Summary

Resume-at-line can recover an older feed rate when the active modal `F` word appears by itself or in tightly packed G-code. The recovered feed is emitted before the recovery Z movement, so an incorrect retained value can affect that move.

On `develop` at `2ea66c10e96d88013a04441134d3d4f8e8c88510`, `_find_last_feed_rate()` only extracts an `F` word from lines it first recognizes as `G1`, `G2`, or `G3`: [feed-recovery implementation](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/Controller.py#L847-L935). The controller then emits that recovered feed before the Z recovery move: [recovery sequencing](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/Controller.py#L1200-L1210).

`F` is modal independently of whether the same line contains a motion command; LinuxCNC lists it among the other modal codes: [G-code quick reference](https://linuxcnc.org/docs/html/gcode.html).

## Minimal reproduction

```gcode
G1 X0 F100
F275
G0 X1
G1 X2
```

Resume at line 4.

Expected:

- Feed recovery returns `275`.
- When a Z position is available, the preview emits `buffer G1 F275` before its recovery Z move.

Actual:

- Feed recovery skips the standalone `F275` and returns the older `F100`.

A tightly packed line such as `X10Y10G1F425` is also missed by the current motion-prefix heuristic.

## Impact

The effect is state-dependent. If the effective feed was set on a line the current scanner misses and the machine retains a different or unknown feed, the recovery Z move can use the wrong feed. If retained feed already matches, no behavior difference is expected. Resume should explicitly reconstruct the program's most recent modal `F` value.

## Acceptance criteria

- Recover the most recent valid `F` word before the resume line, whether standalone, spaced, or tightly packed.
- Ignore `F` words in semicolon and parenthesized comments.
- Emit the recovered feed before the recovery Z move.
- Cover both standalone and tightly packed forms with regression tests.

righteousgambit:
Closes #669

## Summary

Recover modal feed words independently of motion-command formatting so resume-at-line can restore standalone and tightly packed `F` values before recovery movement.

## Changes

- Scan prior lines with the controller's existing G-code tokenizer instead of requiring a recognized `G1`/`G2`/`G3` prefix.
- Select the most recent valid `F` token while continuing to ignore comments.
- Add unit coverage for standalone `F`, tightly packed G-code, comments, and recovery-command ordering.

## Why

Feed is modal independently of a motion word. The previous parser could skip a valid `F275` line or a packed `X10Y10G1F425` line, then restore an older feed before the resume recovery Z move.

The motion risk is conditional on retained controller state; the fix makes the recovery value explicit and deterministic.

## Validation

- Regression against the pinned `develop` implementation: `3 failed, 2 passed`; standalone and packed feed values incorrectly resolved to the older `100`, and the preview omitted `buffer G1 F275`.
- With this change, using the same exact-ref sparse harness: `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=/tmp/carvera-test-stubs:. pytest -q tests/unit/test_resume_feed_rate.py` -> `5 passed`.

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`7df0e9867548`](https://github.com/righteousgambit/Carvera_Controller/commit/7df0e9867548884728170818e3ee34d873eaecae), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run pytest -q tests/unit/test_resume_feed_rate.py
```

Result: **5 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..7df0e9867548884728170818e3ee34d873eaecae`: **passed**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Enhancement: Advanced TLO Calibration option. Here you can set the offset to use from tool setter, and/or the number of repeat probings to use
 - Enhancement: Added a warning popup if the controller version is lower than the firmware. This is not a supported config
 - Enhancement: Added Auto Ext. Out toggle to spindle dropdown and Config and Run screen. This can be used to automatically run a vacuum or compressor when the spindle is running
+- Fixed: Resume-at-line restores feed rates from standalone and tightly packed F words before recovery moves
 - Fixed: Restore Keyboard Jogging state after Probing Popup is closed
 - Fixed: Repeated firmware checks now happen just once
 - Fixed: UI widget updates from the SerialMonitor() now dispatched via the main thread. This should reduce the number of RecycleView related crashes
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -882,52 +882,16 @@ class Controller:
             if start_line < 1 or start_line > len(lines):
                 return None
 
-            # Search backwards for G1/G01/G2/G02/G3/G03 commands
+            # F is modal independently of a motion command, so the most recent
+            # feed word may be on a standalone or tightly packed G-code line.
             for i in range(start_line - 2, -1, -1):
-                original_line = lines[i]
-                line = original_line.strip()
-
-                # Remove comments using string methods
-                if ";" in line:
-                    line = line[: line.index(";")]
-                # Remove parentheses comments using string methods
-                while "(" in line and ")" in line:
-                    start_paren = line.index("(")
-                    end_paren = line.index(")", start_paren)
-                    line = line[:start_paren] + line[end_paren + 1 :]
-
-                line = line.strip()
-                if not line:
-                    continue
-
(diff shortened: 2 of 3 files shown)


SergeBakharev:
I didn't realise that a bare Feed (F) was valid with the gcode parser in our firmware, but it is, and so this issue is valid. The fix looks good to me.

Merging

PR #679, opened 2026-07-13
Fix M03 spindle-speed recovery for resume-at-line

### LINKED ISSUE #670: Resume-at-line spindle recovery ignores zero-padded M03 commands
## Summary

Resume-at-line recognizes zero-padded `M03` as spindle-on state, but its spindle-speed helper only searches for the literal form `M3`. An interrupted program using `M03 S...` can therefore resume with a bare `M3` instead of explicitly restoring its programmed spindle speed.

The mismatch exists on `develop` at `2ea66c10e96d88013a04441134d3d4f8e8c88510`: the speed helper uses `M3(?![0-9])` in duplicated scans ([spindle-speed recovery](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/Controller.py#L742-L844)), while the general command matcher normalizes numeric command values and therefore recognizes `M03` as `M3` ([token matching](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/Controller.py#L938-L971)). When mode recovery succeeds but speed recovery returns `None`, the preview emits bare `M3`: [resume spindle recovery](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/Controller.py#L1161-L1179).

The Community Firmware only updates the stored spindle speed when an `S` argument is present: [firmware spindle handling](https://github.com/Carvera-Community/Carvera_Community_Firmware/blob/feab653def96a959e695049a4f828aeddbcd5547/src/modules/tools/spindle/SpindleControl.cpp#L44-L68).

This is related to but distinct from [#528](https://github.com/Carvera-Community/Carvera_Controller/issues/528), which tracked missing resume metadata when the local cached file was unavailable. This reproduction supplies the G-code lines and specifically isolates zero-padded `M03` speed parsing.

## Minimal reproduction

```gcode
M03 S12000
G1 X1 F100
G1 X2
```

Preview a resume at line 3.

Expected:

- The preview contains `buffer M3 S12000`.

Actual:

- Spindle mode recovery recognizes `M03`, but speed recovery returns no value.
- The preview contains a bare `buffer M3`.

## Impact

This does not imply a deterministic dead spindle: the firmware can retain an earlier speed when bare `M3` is received. The resumed RPM can nevertheless be stale, wrong, or zero depending on retained state. Explicitly restoring the `S` value from `M03 S...` removes that state dependency.

## Acceptance criteria

- Recover spindle speed from both `M3` and zero-padded `M03`.
- Support `S` before or after the command and tightly packed tokens.
- Do not misclassify other commands such as `M030`.
- Ignore command-like text in comments.
- Cover helper behavior and the emitted resume command with regression tests.

righteousgambit:
Closes #670

## Summary

Use the controller's existing tokenizer and normalized command matcher for resume-at-line spindle-speed recovery, allowing zero-padded `M03` commands to restore their programmed `S` value.

## Changes

- Add a focused line parser that recognizes normalized `M3`/`M03` tokens and extracts the last valid `S` token.
- Reuse that parser for both backward-search passes instead of duplicating literal `M3` regular expressions.
- Add tests for zero-padded, tightly packed, inherited-speed, comment, and `M030` boundary cases plus the final resume preview.

## Why

The general resume logic already recognizes `M03` as spindle-on, but the speed helper did not. That mismatch caused the final recovery command to omit the programmed RPM and rely on retained firmware state.

Bare `M3` is not guaranteed to stop the spindle; it can retain an earlier speed. The correctness and safety concern is nondeterministic stale, wrong, or zero RPM depending on state. This change explicitly restores the file's programmed value.

## Validation

- Regression against the pinned `develop` implementation: `5 failed, 1 passed`; zero-padded cases returned `None`, and the preview emitted bare `buffer M3` instead of `buffer M3 S12000`.
- With this change, using the same exact-ref sparse harness: `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=/tmp/carvera-test-stubs:. pytest -q tests/unit/test_resume_m03_speed.py` -> `6 passed`.

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`e055b6bfe50c`](https://github.com/righteousgambit/Carvera_Controller/commit/e055b6bfe50c7b34517e6c2b7a0023f4eb2fec2a), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run pytest -q tests/unit/test_resume_m03_speed.py
```

Result: **6 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..e055b6bfe50c7b34517e6c2b7a0023f4eb2fec2a`: **passed**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Enhancement: Advanced TLO Calibration option. Here you can set the offset to use from tool setter, and/or the number of repeat probings to use
 - Enhancement: Added a warning popup if the controller version is lower than the firmware. This is not a supported config
 - Enhancement: Added Auto Ext. Out toggle to spindle dropdown and Config and Run screen. This can be used to automatically run a vacuum or compressor when the spindle is running
+- Fixed: Resume-at-line restores spindle speed from zero-padded M03 commands
 - Fixed: Restore Keyboard Jogging state after Probing Popup is closed
 - Fixed: Repeated firmware checks now happen just once
 - Fixed: UI widget updates from the SerialMonitor() now dispatched via the main thread. This should reduce the number of RecycleView related crashes
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -765,84 +765,46 @@ class Controller:
             if start_line < 1 or start_line > len(lines):
                 return None
 
-            # First, find the most recent M3 command and check if it has S parameter
+            # First, find the most recent M3 command and check if it has S parameter.
+            # Token matching normalizes zero-padded forms such as M03 without
+            # confusing other M-codes such as M30.
             most_recent_m3_line = None
             for i in range(start_line - 2, -1, -1):
-                original_line = lines[i]
-                line = original_line.strip()
-
-                # Remove comments using string methods
-                if ";" in line:
-                    line = line[: line.index(";")]
-                # Remove parentheses comments using string methods
-                while "(" in line and ")" in line:
-                    start_paren = line.index("(")
-                    end_paren = line.index(")", start_paren)
-                    line = line[:start_paren] + line[end_paren + 1 :]
-
-                line = line.strip()
-                if not line:
-                    continue
-
-                # Check if line contains M3 (not M30, M31, etc.)
-                line_upper = line.upper()
-                # Use regex to find M3 that's not part of M30, M31, etc.
-                # Look for M3 that's not followed by a digit (to avoid M30, M31, etc.)
-                # M3 can be preceded by anything (including digits from other commands like S12000).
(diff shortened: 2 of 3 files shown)


SergeBakharev:
Confirming that the issue reported is real, and the fix is still valid. Good find

PR #680, opened 2026-07-13
Reject blank required probing axes

### LINKED ISSUE #671: Blank conditionally required probe axes pass validation and are omitted from M-code
## Controller version

`develop` at `2ea66c10e96d88013a04441134d3d4f8e8c88510`. This is controller-side probing-form validation.

## Reproduction

Create an operation whose X distance is conditionally required, then leave the X input blank:

```python
operation = BoreOperation("Bore X", requires_x=True, requires_y=False, image_path="")
config = {"X": ""}  # whitespace-only has the same result

assert operation.get_missing_config(config) is None
assert operation.generate(config).strip() == "M461"
```

The same presence-only conditional check is used by Bore, Boss, Angle, and all three ProbeTip variants for both X and Y.

## Expected

An empty or whitespace-only value for a conditionally required axis should be returned as the missing field, just like an absent key. Numeric strings including `"0"` should remain valid.

## Actual

The operation-specific validators only check whether the key exists, so `{"X": ""}` passes that conditional requirement. The shared G-code formatter then omits blank values, producing an M-code without the axis the selected operation requires: [blank-value omission](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/probing/operations/OperationsBase.py#L15-L22).

Affected checks:

- [Bore](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/probing/operations/Bore/BoreOperation.py#L29-L37)
- [Boss](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/probing/operations/Boss/BossOperation.py#L29-L37)
- [Angle](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/probing/operations/Angle/AngleOperation.py#L32-L43)
- [ProbeTip Bore](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/probing/operations/ProbeTip/ProbeTipOperation.py#L34-L42), [Boss](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/probing/operations/ProbeTip/ProbeTipOperation.py#L75-L83), and [Anchor](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/probing/operations/ProbeTip/ProbeTipOperation.py#L116-L124)

## Impact

The controller can treat an incomplete probing form as valid and emit an incomplete probing command. Firmware behavior for such a command is a separate concern; this issue is the controller failing to enforce its own selected-operation requirement.

## Regression proof

A 48-case matrix covers six operation classes, both conditional axes, blank and whitespace-only inputs, plus `"0"` and `"12.5"` positive controls. Against the pinned base, all 24 blank/whitespace cases fail while all 24 numeric-string controls pass.

righteousgambit:
Closes #671

## Summary

- treat empty and whitespace-only X/Y values as missing when the selected operation conditionally requires that axis
- update Bore, Boss, Angle, and the three ProbeTip operation variants consistently
- retain valid numeric strings, including `"0"`
- add a matrix regression test and an unreleased changelog entry

This does not change which operations require which axes; it only makes the existing requirement check agree with the G-code formatter's treatment of blank values.

## Validation

- regression suite on the base: **24 failed, 24 passed** (every blank/whitespace case failed; every numeric-string control passed)
- regression suite with this change: **48 passed**
- `ruff check` passed for all changed Python files
- `git diff --check` passed

Targeted command:

```shell
poetry run python -m pytest -q tests/unit/test_probing_required_axes.py
```

## Changelog

`- Fixed: Treat blank conditionally required X/Y probing inputs as missing`

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`74f380c9de46`](https://github.com/righteousgambit/Carvera_Controller/commit/74f380c9de466cc317653de86a361ffc0261f236), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run python -m pytest -q tests/unit/test_probing_required_axes.py
```

Result: **48 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..74f380c9de466cc317653de86a361ffc0261f236`: **passed**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fixed: Treat blank conditionally required X/Y probing inputs as missing
 - Enhancement: Add multi-select to the remote file browser
 - Enhancement: Display error message in halt popup. Requires halt errors to start with "ERROR: " in the firmware
 - Enhancement: Add popup notice when using stock firmware instead of the Community Firmware
--- a/carveracontroller/addons/probing/operations/Angle/AngleOperation.py
+++ b/carveracontroller/addons/probing/operations/Angle/AngleOperation.py
@@ -32,11 +32,11 @@ class AngleOperation(OperationsBase):
     def get_missing_config(self, config: dict[str, float]):
         if self.requires_x:
             definition = AngleParameterDefinitions.XAxisDistance
-            if not definition.code in config:
+            if definition.code not in config or not config[definition.code].strip():
                 return definition
         if self.requires_y:
             definition = AngleParameterDefinitions.YAxisDistance
-            if not definition.code in config:
+            if definition.code not in config or not config[definition.code].strip():
                 return definition
         definition = AngleParameterDefinitions.ProbeDepth
         if not definition.code in config:
--- a/carveracontroller/addons/probing/operations/Bore/BoreOperation.py
+++ b/carveracontroller/addons/probing/operations/Bore/BoreOperation.py
@@ -29,11 +29,11 @@ class BoreOperation(OperationsBase):
     def get_missing_config(self, config: dict[str, float]):
         if self.requires_x:
             definition = BoreParameterDefinitions.XAxisDistance
-            if not definition.code in config:
+            if definition.code not in config or not config[definition.code].strip():
                 return definition
         if self.requires_y:
             definition = BoreParameterDefinitions.YAxisDistance
-            if not definition.code in config:
+            if definition.code not in config or not config[definition.code].strip():
                 return definition
 
         required_definitions = {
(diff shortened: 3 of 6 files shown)


SergeBakharev:
No response from submitter (https://github.com/Carvera-Community/Carvera_Controller/pull/682#issuecomment-4987864661)
The PR looks complete and ready for review. Treating it as such.

SergeBakharev:
Confirm that the issue reported is real, and the fix is still valid.

PR #681, opened 2026-07-13
Sanitize WHB04 spindle display values

### LINKED ISSUE #672: WHB04 display refresh raises on missing or malformed spindle values
## Controller version

`develop` at `2ea66c10e96d88013a04441134d3d4f8e8c88510`. This is controller-side display-state handling; firmware is not changed.

## Reproduction

Run one WHB04 display callback while `CNC.vars` has no `curspindle` entry:

```python
pendant._cnc.vars = {
    "wx": 0.0,
    "wy": 0.0,
    "wz": 0.0,
    "wa": 0.0,
    "curfeed": 0.0,
}
pendant._handle_display_update(daemon)
```

Malformed (`None`, `"invalid"`), non-finite (`nan`, `+inf`, `-inf`), negative, over-range, and finite integers too large for `float` conversion also reproduce an exception or setter rejection. For example, `float(10**10000)` raises `OverflowError` in the sanitizing path.

## Expected

The scheduled display refresh should remain live. Missing, malformed, or non-finite spindle values should display as zero; finite values should be clamped to the WHB04's supported `0..65535` range.

## Actual

The callback indexes `curspindle` directly and passes it unsanitized to a strict setter: [display callback](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/pendant/pendant.py#L170-L184), [setter](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/pendant/whb04.py#L283-L289). The daemon schedules that handler onto the GUI clock from its device loop ([scheduling call](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/pendant/whb04.py#L389-L391), [clock executor](https://github.com/Carvera-Community/Carvera_Controller/blob/2ea66c10e96d88013a04441134d3d4f8e8c88510/carveracontroller/addons/pendant/pendant.py#L103-L104)).

## Impact

The GUI-clock callback raises and the pendant can retain a stale spindle value (and skip the remainder of that display update). Because the callback is scheduled asynchronously, this is not evidence that the HID read loop reconnects, nor does the callback itself command or alter spindle motion.

## Regression proof

An eleven-case test starts the daemon display at a nonzero value and covers missing, `None`, malformed text, negative, over-range, numeric text, `nan`, `+inf`, `-inf`, and huge positive/negative integers. Against the pinned base all eleven cases raise or retain invalid behavior. The previous patch revision passed the first nine but failed both huge-integer cases with `OverflowError`.

righteousgambit:
Closes #672

## Summary

- read `curspindle` with a zero default
- normalize missing, malformed, and non-finite values to zero
- clamp finite values, including integers outside `float`'s range, to the WHB04 setter's `0..65535` range
- add an eleven-case regression matrix that starts from a nonzero display value
- add the unreleased changelog entry

The change only sanitizes what is shown on the pendant. It does not write CNC state or issue a spindle command.

## Validation

- regression matrix on the pinned upstream base: **11 failed**
- adversarial test on the previous patch revision: **2 failed, 9 passed**; both huge integers raised `OverflowError`
- regression matrix with this change: **11 passed**
- `ruff check` passed for the source and test
- `git diff --check` passed

Targeted command:

```shell
poetry run python -m pytest -q tests/unit/test_pendant_display.py
```

## Changelog

`- Fixed: Sanitize missing or malformed spindle values before updating the WHB04 display`

## Real branch verification

The repository's locked Poetry environment was installed from `poetry.lock`. The targeted suite passed on real fork commit [`75dc25352e51`](https://github.com/righteousgambit/Carvera_Controller/commit/75dc25352e51dbe834d00378715a6f516e2b19b0), whose sole parent is pinned upstream `develop@2ea66c10e96d88013a04441134d3d4f8e8c88510`.

```shell
poetry run python -m pytest -q tests/unit/test_pendant_display.py
```

Result: **11 passed**.

- Ruff `check`: **passed**
- `git diff --check 2ea66c10e96d88013a04441134d3d4f8e8c88510..75dc25352e51dbe834d00378715a6f516e2b19b0`: **passed**

### DIFF
--- a/tests/unit/test_pendant_display.py
+++ b/tests/unit/test_pendant_display.py
@@ -1,0 +1,49 @@
+"""Tests for defensive pendant display updates."""
+
+from types import SimpleNamespace
+
+import pytest
+
+from carveracontroller.addons.pendant import pendant as pendant_module
+from carveracontroller.addons.pendant import whb04
+
+
+@pytest.mark.parametrize(
+    ("spindle_vars", "expected_speed"),
+    [
+        ({}, 0),
+        ({"curspindle": None}, 0),
+        ({"curspindle": "invalid"}, 0),
+        ({"curspindle": -1}, 0),
+        ({"curspindle": 70000}, 65535),
+        ({"curspindle": "1234.5"}, 1234),
+        ({"curspindle": float("nan")}, 0),
+        ({"curspindle": float("inf")}, 0),
+        ({"curspindle": float("-inf")}, 0),
+        pytest.param({"curspindle": 10**10000}, 65535, id="huge-positive-integer"),
+        pytest.param({"curspindle": -(10**10000)}, 0, id="huge-negative-integer"),
+    ],
+)
+def test_whb04_display_sanitizes_spindle_speed(spindle_vars, expected_speed):
+    pendant = pendant_module.WHB04.__new__(pendant_module.WHB04)
+    pendant._cnc = SimpleNamespace(
+        vars={
+            "wx": 0.0,
+            "wy": 0.0,
+            "wz": 0.0,
+            "wa": 0.0,
+            "curfeed": 0.0,
+            **spindle_vars,
+        }
+    )
+    pendant._controller = SimpleNamespace(
+        jog_mode=0,
+        JOG_MODE_CONTINUOUS=1,
+        JOG_MODE_STEP=0,
+    )
+    daemon = whb04.Daemon()
+    daemon.set_display_spindle_speed(4321)
+
+    pendant._handle_display_update(daemon)
+
+    assert daemon._display_spindle_speed == expected_speed
--- a/.github/workflows/quality.yaml
+++ b/.github/workflows/quality.yaml
@@ -28,16 +28,17 @@ jobs:
           path: .venv
           key: ${{ runner.os }}-quality-py3.12-${{ hashFiles('poetry.lock') }}
 
-      - name: Install Linux test prerequisites
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y xvfb libgl1 libgles2 libmtdev1
-
       - name: Install poetry and dependencies
         uses: ./.github/actions/bootstrap-poetry
         with:
           os: linux
           install-args: --only main,lint,test
+
+      - name: Install Linux prerequisites
+        run: poetry run ./scripts/install_linux_prereqs.sh
+
+      - name: Install Linux test prerequisites
+        run: sudo apt-get install -y xvfb libgl1 libgles2 libmtdev1
 
       - name: Run quality hooks
         run: poetry run pre-commit run --all-files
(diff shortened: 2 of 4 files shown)


SergeBakharev:
Same as https://github.com/Carvera-Community/Carvera_Controller/pull/677#issuecomment-4963932214

righteousgambit:
Addressed in [`1cf1460bac5d`](https://github.com/righteousgambit/Carvera_Controller/commit/1cf1460bac5d3db2089987dff135c2f0d83f706e).

The Quality Gate now runs `poetry run ./scripts/install_linux_prereqs.sh` after Poetry bootstrap, so the Ubuntu runner installs `libhidapi-hidraw0`, and then installs the remaining Xvfb/OpenGL test packages. The resulting workflow blob is byte-for-byte identical to the correction requested and applied on #677.

Local verification on the updated branch:

- pendant regression suite: **11 passed**
- Ruff check and format check: **passed**
- workflow validation (filtering the repository's pre-existing `action.yml` `options` diagnostic): **passed**
- shell syntax and `git diff --check`: **passed**

The exact-head [Quality Gate run](https://github.com/Carvera-Community/Carvera_Controller/actions/runs/29381981940) is currently `action_required` with zero jobs, so it needs upstream approval before the runner can exercise the new prerequisite step.

SergeBakharev:
No response from submitter (https://github.com/Carvera-Community/Carvera_Controller/pull/682#issuecomment-4987864661)
The PR looks complete and ready for review. Treating it as such.

tests/unit/test_pendant_display.py
```diff
@@ -0,0 +1,49 @@
...
+        ({"curspindle": None}, 0),
+        ({"curspindle": "invalid"}, 0),
+        ({"curspindle": -1}, 0),
+        ({"curspindle": 70000}, 65535),
+        ({"curspindle": "1234.5"}, 1234),
+        ({"curspindle": float("nan")}, 0),
+        ({"curspindle": float("inf")}, 0),
+        ({"curspindle": float("-inf")}, 0),
+        pytest.param({"curspindle": 10**10000}, 65535, id="huge-positive-integer"),
+        pytest.param({"curspindle": -(10**10000)}, 0, id="huge-negative-integer"),
```
SergeBakharev, line 24: Normal curspindle values come from float(), so these ints aren't really valid. But I guess since we don't having type hints, this doesn't hurt to be a little bit adversarial in testing.

SergeBakharev:
Thanks for adding the prereqs for the testing CI step.

Confirming that the issue reported is real, and the fix is still valid. I've pulled latest, and resolve the changlog merge conflict.

Good to merge

Issue #419, opened 2025-11-25
Improve tool-change popup to show Fusion CAM tool metadata and operation name


prdoring:
# Improve tool-change popup to show Fusion CAM tool metadata and operation name

## Summary
Enhance the tool-change confirmation popup to include Fusion CAM–provided tool metadata (e.g., diameter, corner radius, taper) and the current operation header (e.g., `(Back Carve)`), shown alongside the existing "Please change to tool X" message.

## Scope / Note
This proposal is specific to Fusion 360 (and similar CAMs that embed parenthetical header comments) G-code exports. I haven't audited other CAM formats yet — behavior and header formats can vary between CAMs, so this change intentionally targets Fusion-style headers first.

## Motivation
Fusion-exported G-code commonly includes descriptive headers such as:

```
(T1 D=3.175 CR=0 TAPER=30deg - chamfer mill)
(Back Carve)
```

Showing only the tool number (e.g., "Please change to tool: 1") is ambiguous for operators who rely on CAM-provided names and parameters to identify the correct cutter. Including the human-readable name and key parameters reduces errors and speeds up manual changes.

## Proposed behavior
- When the machine reports a tool change (runtime bracketed status supplies `target_tool`), the popup should display:
  - Primary line (existing): `Please change to tool: 1`
  - Secondary line: `Tool info: Chamfer Mill (D=3.175 CR=0 TAPER=30deg)` — omit `ZMIN`
  - Secondary line: `Operation: Back Carve` (when available and relevant)
- If no description or operation header is available, fall back to the current minimal message.
- Prefer metadata from a loaded local file; when playing remote files, use a cached copy in `temp_dir` if available.
- Parsing should handle common Fusion header variations; parameters parsed as key=value (D, CR, TAPER, etc.) and presented in parentheses after the human name.

## Technical notes / Implementation hints
- Candidate parsing locations (priority order):
  1. `self.lines` (loaded file buffer)
  2. `app.selected_local_filename` (local path)
  3. cached remote copy in `self.temp_dir` (remote playback)
- Suggested extractor behavior for Fusion-style headers:
  - Look for parenthetical header lines beginning with `Tn` (e.g., `(T2 D=4.0 - flat end mill)`).
  - Extract key=value pairs via regex, e.g. `([A-Za-z]+)=([^\s\)]+)`.
  - Omit `ZMIN` (case-insensitive).
[...]

PR #685, opened 2026-07-15
Load and show tool geometry / info in the 3D viewer

Lyrkan:
Related to #419 

I'll add some more details here when I get the time to do so but this PR adds a way to extract tools information from the comments added by CAMs  post-processors in order to display them and have more accurate representations in the 3D viewer.

The idea is that for each supported post-processor we would just have to need to add a specific parser in `
carveracontroller/addons/tool_visualization/parsers/` that detects tool descriptions comments and convert them to a common `ToolDefinition` object.

For now I only created one for Fusion as a proof-of-concept (edit: also added something for Makera Studio) using the current post-processor format (which will probably need to be improved to replace localized string by constants, and maybe add a few extra parameters if we can such as thread pitch/diameter). @faecorrigan Since you worked a lot on post-processors does it sounds like a good way to handle that? Do all of them allow to expose at least some basic info?

The tools icons are only displayed if a parser detected at least one tool description and if there is enough space for them. The images could probably be improved since I didn't spend a lot of time on them.

I also added some tooltips on those buttons to display more details about each tool.

Some screenshots:

**Tooltip on hover:**
image

**Manual tool change modal:**
image

**Ball endmill:**
image

**Chamfer (or any kind of pointy) endmill:**
image

**Flat-end endmill:**
image

**Radius endmill:**
image

**Thread endmill:**
image

**Lollipop endmill:**
image

**Tapered endmill:**
image

**Using Makera Studio format:**
image

**Using FreeCAD:**
image

**Fallback when no tool definition was found:**
image


SergeBakharev:
I like the approach.

It would be good to have the tool description in the popup for manual tool loading on the non-atc machines

Lyrkan:
> It would be good to have the tool description in the popup for manual tool loading on the non-atc machines

Yep, that's a good idea, I implemented a first version of it.

I also added a parser for Makera Studio based on the few samples of gcode files I've seen posted on Discord (so it's probably wrong on some things but hard to verify without access to it).

They seem to contain more info than what we currently have for Fusion (shank diameter and flute length for instance), which makes it a bit easier since we don't have to guess or make up values.

Lyrkan:
Added support for the flute/shoulder/shank length transitions in the generated mesh.

I also made it so the flute part of the mesh is now colored differently (gold) than the rest of the tool (which is now gray instead of blue).

Of course for Fusion most of those dimensions come from heuristics since we only know a few values (diameter, corner radius and taper angle). With the Makera Studio format it's working quite well though:

tools-viz-makera-v3

Lyrkan:
I updated the icons with better images (see updated screenshots in the original post).

Disclosure on how I generated them:
* got some reference pictures from tools sold on Ali
* cleaned them up a little bit and asked Gemini to generate a stylized/simplified version of them
* had to do a lot more cleanup then since the LLM didn't fully respect input geometries and added some artifacts
* cropped the resulting image into multiple icons, each one in two versions: one for the buttons (zoomed on the tip) and one for the tooltips

I'm sure they could be way better but I'm really bad at that kind of thing, so if anyone wanna give a shot at it feel free to do so!

Lyrkan:
Added support for a few additional info I added to the Fusion post in the following PR: https://github.com/Carvera-Community/Carvera_Community_Profiles/pull/33

image

Lyrkan:
Got my hands on Makera Studio so the related parser should now be mostly fine.

Something a bit weird though: in the sample I initially used every tool attribute was properly defined, for instance:

```
;@MKR|TOOL|number=1|id=111011203812|name=3.175*2*6mm Flat End|type=Flat End|handlediameter=3.175|sticklength=12|shoulderlength=8|flutelength=6|diameter=2|tipdiameter=2|cornerradius=0|angle=0|halfAngle=0
```

But from what I've seen some of them are actually missing from the default library included in Studio. For instance the `sticklength` is always 0 in the following block (and I don't even see where it could be configured in the UI...)  and most of the time `shoulderlength` = `flutelength`, which means that only the "gold" part will be showing for them.  Some values are also a bit weird (`shoulderlength` < `flutelength`, `tipdiameter` > 0 on a chamfer mill, etc.).

```
;@MKR|TOOL|number=1|id=112111203808|name=3.175*2*8mm Flat End(Metal)|type=Flat End|handlediameter=3.175|sticklength=0|shoulderlength=8|flutelength=8|diameter=2|tipdiameter=2|cornerradius=0|angle=0|halfAngle=0
;@MKR|TOOL|number=2|id=142212313812|name=3.175*12mm Drill|type=Drill|handlediameter=3.175|sticklength=0|shoulderlength=12|flutelength=12|diameter=3.175|tipdiameter=3.175|cornerradius=0|angle=0|halfAngle=0
;@MKR|TOOL|number=3|id=152111243809|name=3.175*M3*9mm Thread|type=Thread|handlediameter=3.175|sticklength=0|shoulderlength=9|flutelength=10|diameter=2.42|tipdiameter=0|cornerradius=0|angle=0|halfAngle=0
;@MKR|TOOL|number=4|id=122111013890|name=3.175mm*90º Chamfer|type=V-Bit|handlediameter=3.175|sticklength=0|shoulderlength=0|flutelength=1.6|diameter=3.175|tipdiameter=0.1|cornerradius=0|angle=90|halfAngle=0
;@MKR|TOOL|number=5|id=112131605017|name=6*17mm Flat End(Metal)|type=Flat End|handlediameter=6|sticklength=0|shoulderlength=17|flutelength=17|diameter=6|tipdiameter=6|cornerradius=0|angle=0|halfAngle=0
```

Lyrkan:
Added a new parser dedicated to FreeCAD that uses the post processor improvements submitted in https://github.com/Carvera-Community/Carvera_Community_Profiles/pull/34

image

Lyrkan:
Since I was still unhappy with the icons I decided to try another approach and generate them dynamically using the same profiles used by the 3D viewer (which also allows to show the real geometry)... and I'm quite pleased with the result (see updated screenshots in my original post)!

I also improved the tooltip quite a lot so it does not just display a wall of random values anymore.

I wouldn't say this is perfect yet but that should be good enough for now!

faecorrigan:
I also like the new icons matching the tool as it shows up in the previewer, very clean. The helixes were a bit too noisy for the size of the icon

I am working through the post processors this weekend and have merged the code for the new tool headers.
I have a couple more bugs to track down, but I should be good to release new posts this week.

SergeBakharev:
I've tested some of my existing files including manual toolchanges. Looks great to me!

@faecorrigan If all good on the post-processor change side, happy to merge this one

SergeBakharev:
I see you merged all the post-processor changes, merging this too then

Issue #560, opened 2026-04-12
Auto-reconnect support for USB-Serial


SergeBakharev:
Auto-reconnect is very convenient, it would be great to extend the functionality to usb-serial

Issue #491, opened 2026-02-01
Backport new Comms protocol


SergeBakharev:
We should backport the new comms protocol from the newer Makera firmware/controller. It should however continue to be compatible with the older comms standard

SergeBakharev:
I plan to write compatibility/support article for the documentation to outline it better.

While the Community Controller has been developed with compatibility with Makera Firmware in mind, the latest stable release 2.0.0 is not compatible with Makera Firmware >= 1.0.4. Because as you correctly guessed the comms change. We will have the new comms protocol merged in soon, we have been busy finishing up the v2.1.0 release of the community software

PR #686, opened 2026-07-18
Comms uplift

SergeBakharev:
Changes:
- Enhancement: Autodetect Smoothie vs Makera communication protocol on connect and use it for the session
- Enhancement: Show a connecting progress popup while opening a USB device
- Enhancement: Log connect and manual disconnect with the connection method and address
- Enhancement: Add a "Network..." option under Scan Wi-Fi in the connection dropdown to enter a machine network address
- Enhancement: Reconnect supports USB as well as WiFi. Configure preferred method for app-launch auto-connect in Controller settings
- Change: Auto-connect on app launch now is only performed if auto-reconnect is enabled
- Change: Reconnect uses the last successful connection method. On fresh app launch it uses the configured preferred connection method
- Change: USB devices in connection dropdown are filtered to only show devices specifically with the FTDI chip found on the Makera machines
- Change: USB devices are now selected and stored by stable VID:PID:serial identity (instead of generic COM path)

Resolves: #560 #491 

Screenshot 2026-07-18 at 7 22 59 pm
Screenshot 2026-07-18 at 7 22 34 pm
Screenshot 2026-07-18 at 7 21 27 pm

I tested with the Makera native firmware on the machine, as well as both modes from the new switchable code from [PR343](https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/343)

It was even able to update the machine firmware from running 1.0.6 Makera to our Community version. So I'm pretty confident in all that.

Where I'm less confident is in the protocol abstraction, so in future we can implement our own thing. I'm having difficulty imagining what that might look like, probably doesn't have to be perfect today. But I would still appreciate those comments.

As part of the USB changes I made it filter the USB device selection to just the FTDI chip, and record the device serial to config for reconnect. This works well on my MacOS machine, if someone can help test this on Windows that would be awesome

### DIFF
--- a/carveracontroller/protocols/registry.py
+++ b/carveracontroller/protocols/registry.py
@@ -1,0 +1,37 @@
+"""Protocol registry — register new protocols here to make them selectable."""
+
+from __future__ import annotations
+
+from .base import CommunicationProtocol
+from .makera import MakeraProtocol
+from .smoothie import SmoothieProtocol
+
+# Map of protocol name -> implementation class.
+# To add a third protocol: implement CommunicationProtocol and register it here.
+PROTOCOL_CLASSES: dict[str, type[CommunicationProtocol]] = {
+    "smoothie": SmoothieProtocol,
+    "makera": MakeraProtocol,
+}
+
+DEFAULT_PROTOCOL = "smoothie"
+
+
+def available_protocols() -> tuple[str, ...]:
+    return tuple(PROTOCOL_CLASSES.keys())
+
+
+def create_protocol(name: str) -> CommunicationProtocol:
+    """Instantiate a protocol by registered name."""
+    try:
+        cls = PROTOCOL_CLASSES[name]
+    except KeyError as exc:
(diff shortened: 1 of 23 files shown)


SergeBakharev:
I'm noticing a race condition for the continuous jogging keep alive. Fixing....

carveracontroller/protocols/registry.py
```diff
@@ -0,0 +1,37 @@
...
+from .smoothie import SmoothieProtocol
+
+# Map of protocol name -> implementation class.
+# To add a third protocol: implement CommunicationProtocol and register it here.
+PROTOCOL_CLASSES: dict[str, type[CommunicationProtocol]] = {
+    "smoothie": SmoothieProtocol,
+    "makera": MakeraProtocol,
+}
+
+DEFAULT_PROTOCOL = "smoothie"
```
SergeBakharev: Should change this default to Makera if that's what we are going with in https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/343

PR #694, opened 2026-07-22
Z1 camera support

MichaelLevAstro:
Work in Progress! Untested and dirty first impl


SergeBakharev:
I've marked this PR as draft to reflect that it is WiP

carveracontroller/data/video.png
SergeBakharev, line 1: Did you draw this? If not what is the source?

MichaelLevAstro: No, AI just took a licensed icon, will redirect to the already existing eye icon.
Unless you want to add a new unlicensed video icon?

carveracontroller/addons/camera/sources/base.py
SergeBakharev, line 1: I like that the camera sources are pluggable, however it looks like irrespective of the implementation it requires the source to be providing JPEG frames. So really it's only the transport that is pluggable here. Many cameras are H264 or even raw. 

The third variable is controls, not all cameras have the same controls available. I have even seen an Carvera Air owner on FB with a small pan/tilt camera attached to the spindle.

I suggest investigating if there is a library that could help abstract all this complexity, instead of having to re-write the wheel. If you come up short, then the lets look to include all three elements: Transport, Codec, Controls.

To be honest, I didn't expect you to go straight for the multi-source feature. Maybe let's walk first with explicit support only for the Z1 Camera, and once that is merged and we have a better idea of how things fit together we can look to abstract the implementation to support other types of cameras.

carveracontroller/data/video.png
SergeBakharev: I see. I suggest using https://pictogrammers.com/library/mdi/icon/camera/ as we are already using other icons from this source, and the acknowledgement to them is already in NOTICE. 

How much of this code is a copy of the Makera implementation? If any of it is, we should find a way to properly attribute them. I need to do some thinking there, as we started the Community Controller project before Makera had officially made their source public. Tracking this as #696 since this needs to be done irrespective.

carveracontroller/main.py
```diff
@@ -7169,6 +7192,34 @@ def gcode_play_toggle(self):
...
+    def _detect_camera(self):
+        endpoint = find_camera(self.controller.connection_address.split(":")[0])
+        Clock.schedule_once(partial(self._on_camera_detected, endpoint), 0)
+
+    # -----------------------------------------------------------------------
+    def _on_camera_detected(self, endpoint, *args):
+        if not self.camera_checked:
+            return
+        self.camera_endpoint = endpoint
+        App.get_running_app().supports_camera = endpoint is not None
```
SergeBakharev: There is a race condition here. Fast disconnect→reconnect can apply a stale probe. Disconnect clears camera_checked, reconnect sets it True and starts a new probe; an older probe completing later still sees camera_checked and can overwrite camera_endpoint / supports_camera.

carveracontroller/addons/camera/CameraStream.py
```diff
@@ -0,0 +1,112 @@
...
+        finally:
+            if source is not None:
+                source.close()
+            if self._session == session:
+                self._source = None
+                self._streaming = False
+                self._notify(self._on_streaming, False)
+
+    def _notify(self, callback, value):
+        Clock.schedule_once(lambda _dt: callback(value), 0)
```
SergeBakharev: This potentially will queue multiple frames to be shown. If the host can't keep up then the queue of frames will grow and grow. 

See the pattern used in `GcodePlaySlider._update_line_highlighting()` for how to keep only one update scheduled at a time.

carveracontroller/addons/camera/CameraView.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    texture = ObjectProperty(None, allownone=True)
+    brightness = NumericProperty(ADJUST_DEFAULT)
+    contrast = NumericProperty(ADJUST_DEFAULT)
+    gamma = NumericProperty(ADJUST_DEFAULT)
+
+    def __init__(self, **kwargs):
+        self.canvas = RenderContext(use_parent_projection=True, use_parent_modelview=True)
+        self.canvas.shader.fs = _FRAGMENT_SHADER
```
SergeBakharev, line 47: It would be good to test this custom shader on Android/iOS since those mobile systems can have slightly different OpenGL (GL vs GLES).

Actually, why re-invent the wheel. Why not use the Kivy native [Camera](https://kivy.org/doc/stable/api-kivy.core.camera.html) or [Video](https://kivy.org/doc/stable/api-kivy.uix.video.html) widgets?

This might also help address https://github.com/Carvera-Community/Carvera_Controller/pull/694/changes#r3649685897 since it appears to use gstreamer...

carveracontroller/addons/camera/sources/base.py
MichaelLevAstro: Fair points, and agreed. I've dropped the source abstraction and gone back to explicit Z1 support only.

`sources/`, `endpoints.py` and `CameraStream.py` are gone, replaced by a single `Z1Camera.py` (a `has_camera()` probe plus the session class) on top of `websocket.py`, which was the one piece worth keeping. The MJPEG-over-HTTP source went with them — it was written for a camera nobody has, and on a Z1 port 81 is the machine's JSON control channel anyway.

Happy to revisit transport/codec/controls, or a library that already covers all three, once there's a second camera to design against.

carveracontroller/main.py
MichaelLevAstro: Good catch. The probe now carries the id of the connection that started it, and the result is thrown away if a disconnect or a newer probe has since superseded it. `camera_checked` on its own couldn't express that, exactly as you describe.

carveracontroller/addons/camera/CameraStream.py
MichaelLevAstro: Fixed. Only one update is scheduled at a time now, and it always draws the newest frame, so a host that can't keep up drops frames instead of building a backlog.

Same idea as `_update_line_highlighting()`, minus the unschedule/reschedule — the already-pending update picks up the latest frame on its own.

carveracontroller/addons/camera/CameraView.py
MichaelLevAstro: I looked at both. Kivy's `Camera` only wraps local capture devices (the opencv/picamera/android providers), so it can't read a network stream. `Video` takes a URI through ffpyplayer or gstreamer, so it can't speak WebSocket, and it would pull a fairly heavy recipe into the Android and iOS builds. Neither helps with the abstraction question either — transport was never the hard part there.

On the shader: custom GLSL isn't new here. `GcodeViewer.py` already ships five `$HEADER$` shaders from `carveracontroller/shaders/`, and those run on Android and iOS today. This one only uses `texture2D`, `pow` and `clamp`, all core GLSL ES 1.00, and `$HEADER$` supplies the `GL_ES` precision block, so the GL vs GLES risk is the same one the toolpath view already carries. That said, it's untested on a device on my side, so worth a look if someone has one handy.

carveracontroller/data/video.png
MichaelLevAstro: Switched to the MDI camera icon — `data/camera.png` rasterized from `svg/camera.svg`, with its own NOTICE entry next to axis-arrow.png.

On the Makera question: as far as I can tell there's nothing to have copied. `MakeraInc/CarveraController` has no camera implementation at all — 13 Python files, no websocket or video code, just a leftover `video_connected` property in `makera.py` and `src/data/video.png`. The camera lives in their separate Z1 application, which only ships compiled.

The one thing that genuinely was a copy is that icon: the `data/video.png` I removed in 8ba50bd was byte-identical to `src/data/video.png` in their repo (same blob, `3afc6ae`). It's gone as of this branch, though the blob is still reachable through earlier commits here.

What did come from their Z1 app is the protocol — port 82, `/ws_video`, a text `start_stream`, then one JPEG per binary message. I worked that out by watching the app, then confirmed it against the machine's own web UI, which does the identical handshake, so it's really the machine's interface rather than their code. The implementation is written from scratch: hand-rolled RFC 6455 client, Kivy `CoreImage` for the decode, our own session handling.

SergeBakharev: Awesome great work!

carveracontroller/addons/camera/CameraView.py
SergeBakharev: I had no issue with the use of the shader (I think it's the right way of doing this), just that I was that previously this had been a source of cross-platform inconsistency. Thanks for comparing what actually is available in GL vs GLES.

Fantastic work investigating the Camera/Video widgets, it's a shame there isn't a nice one-size-fits-all widget solution here, but I'm not surprised video is notorious for this. This makes me more confident in this approach

CHANGELOG.md
```diff
@@ -57,6 +58,8 @@
 - Fixed: Reject downloads whose content does not match the machine-provided MD5. Skip MD5 check when none is available, and defer .lz checks until after decompress
 - Fixed: Ensure complete XMODEM packets are written over Wi-Fi
 - Fixed: Confirm popup content now scrolls and sizes to its text
+- Fixed: Dragging a slider that floats over the gcode viewer also orbited or panned the view behind it
+- Fixed: Every download failed its MD5 check on firmware that answers md5sum with a fixed placeholder instead of a digest, as stock Z1 firmware does
```
SergeBakharev: Clearer text

```suggestion
- Fixed: On the Makera Z1 firmware every download returns same placeholder MD5 hash instead of a digest failing the MD5 check
```

Issue #710, opened 2026-08-05
Features that send multiple commands in rapid succession cause Controller disconnect


SergeBakharev:
Features such as the CMM Workbench and resume-at-line send multiple commands in quick succession. If the machine is running in Makera communication protocol mode (as is default in 2.2.0) this causes a connection drop due to the firmware issue https://github.com/Carvera-Community/Carvera_Community_Firmware/issues/400

As a workaround for now please switch the machine to using the `Smoothie` protocol either temporarily using `M485.1` or via persistent config `config-set sd protocol smoothie`

SergeBakharev:
Closing since we resolved it in firmware https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/407

PR #713, opened 2026-08-06
CMM Workbench: Fix some timeout issues and improve unexpected states detection

Lyrkan:
Related to https://github.com/Carvera-Community/Carvera_Controller/issues/710

The current version of the CMM tool sends a lot of commands at the same time (setup + probing command + M118 calls to retrieve the result). It worked fine with the smoothie protocol but the new one only supports one deferred command at a time on Wifi and blocks all incoming packets until that slot is free.

This basically means that if you send a commands that takes some time to process (here a probing command) followed by another one (here a M118), the latter will hold the deferred spot until the first command has been fully processed. It will then block all incoming packets, including the status query requests, and cause the controller to timeout after 5s.

This PR updates the CMM tool in order to send commands in 3 different waves: setup, then probing, then M118 calls (the first two could probably be sent together but I felt it was cleaner this way). This requires to wait for the machine to go back to and idle state before starting the next step.

This method also allows to detect the machine going into unexpected state and directly abort the current operation instead of having to wait 30s for it to timeout.

This is not perfect though:
* if for some reason the controller decides to send another command (other than the query status) while the probing command is being processed it will probably also cause a timeout. But I don't see why it would happen since the controller isn't supposed to send random commands in the background (apart from the status query)
* if assumes the machine will go idle => run => idle => run => idle. I added some grace period so that if it misses a "run" it is able to considers the next "idle" state as being ok, but that could cause false positives if there is a really big lag (> 3s) between the moment a command is sent and the moment the machine goes into its "run" state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+[unreleased]
+- Fixed: Fix some timeout issues with the CMM workbench when using the Makera protocol and improve unexpected states detection
+
 [2.2.0-RC1]
 - Enhancement: Read tool definitions from post-processor outputs and use them in the G-code viewer
 - Enhancement: Add multi-select to the remote file browser
--- a/carveracontroller/addons/cmm_workbench/core/gcode.py
+++ b/carveracontroller/addons/cmm_workbench/core/gcode.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import re
 from collections.abc import Callable, Sequence
+from dataclasses import dataclass
 
 from carveracontroller.Controller import Controller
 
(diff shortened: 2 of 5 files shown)


SergeBakharev:
I prefer to explore if we can add more "pending" commands to make it more or parity with the smoothie comms https://github.com/Carvera-Community/Carvera_Community_Firmware/issues/400 I feel like this would be a better solution.

Lets keep this PR open until we determine how far we can take the firmware.

Lyrkan:
> I prefer to explore if we can add more "pending" commands to make it more or parity with the smoothie comms [Carvera-Community/Carvera_Community_Firmware#400](https://github.com/Carvera-Community/Carvera_Community_Firmware/issues/400) I feel like this would be a better solution.
> 
> Lets keep this PR open until we determine how far we can take the firmware.

Fair enough, not sure how easy it will be to fix it in the firmware without hitting other pitfalls though :/

If we were to abandon this PR I still think that the "unexpected state check" part should be implemented in a new one since waiting for the CMM tool to timeout when a probing operation fails is a bit annoying.

Lyrkan:
Closing this one in favor of #729 since https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/407 should fix the commands queue.

PR #722, opened 2026-08-08
Add Brazilian Portuguese (pt-BR) translation

oliveiracelso:
Closes #720

## What

Adds a complete Brazilian Portuguese (pt-BR) translation:

- `carveracontroller/locales/pt-BR/LC_MESSAGES/pt-BR.po` with all 424 strings from `messages.pot` translated (no fuzzy/empty entries). The compiled `.mo` is not committed, per `.gitignore`.
- Registers `pt-BR` in `translation.LANGS`, displayed as **Português (Brasil)** in the language selector.
- Adds `pt-BR` to `LANGUAGES` in `scripts/update_translations.py` so `msgmerge`/`msgfmt` maintenance covers it.

## Translation approach

- Terminology follows the original Makera Controller `pt` locale where strings overlap (247 entries): *Sonda* for probe, *Offset* kept as-is, *Avanço* for feed, etc. — users migrating from the stock controller see familiar wording.
- The 176 community-specific strings (probing suite, WCS management, pendant, MDI, jog controls, tooltips) were translated by a native Brazilian Portuguese speaker.
- Whitespace-sensitive msgids (leading/trailing spaces, embedded `\n`) preserved exactly.

## Testing

- `gettext.translation("pt-BR", ...)` loads the compiled catalog and returns the expected strings.
- App launched with `carvera.language = pt-BR`: UI renders in Portuguese, no missing-glyph or layout issues observed in the main screens, dropdowns and popups.
- Locale auto-detection picks pt-BR automatically on `pt_BR` systems via the existing two-letter prefix match in `translation.init()`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #723, opened 2026-08-08
Keep MDI input within the machine command buffer limits

### LINKED ISSUE #718: MDI Input allows for more content than buffer on machine
Similar but distinctly different to https://github.com/Carvera-Community/Carvera_Controller/issues/710

Even in `smoothie` comms mode, it's possible for long MDI terminal input to:
1. Get mangled with the 200ms status query
2. Be longer than the machine's input buffer resulting in the part of the input being truncated.

oliveiracelso:
Fixes #718 (controller side).

## Firmware limits this enforces

Digging through `Carvera_Community_Firmware` for the actual bounds:

| Transport / protocol | Limit | Failure mode today |
|---|---|---|
| USB + WiFi, smoothie text mode | `RingBuffer<char, 256>` receive buffer (`SerialConsole.h`, `WifiProvider.h`) | chars beyond capacity are lost → truncated command |
| USB, makera framed mode | `MAKERA_CMD_MAX_LEN = 256` per queued `CTRL_MULTI` payload, queue depth 4 (`SerialConsole.cpp`) | oversized payload is rejected by `makera_cmd_queue_push` → command silently dropped |
| any, makera framed mode | whole `CTRL_MULTI` payload dispatched as **one** `SerialMessage` regardless of embedded newlines (`SerialConsole::on_main_loop`, `WifiProvider::on_main_loop`) | multi-line MDI text lumped into a single console line |

On the controller side there were two additional mangling vectors:

- `WIFIStream.send()` used `socket.send()` and ignored the return value — a partial send silently dropped the tail of the command.
- The 200 ms status poll (`streamIO` thread) and command writes (UI thread) share one socket/serial port with no synchronization, so writes could interleave.

## Changes

- **`carveracontroller/machine/mdi.py`** (new, Kivy-free, `mypy --strict` clean): `prepare_mdi_lines()` splits MDI input into individual command lines and rejects lines whose UTF-8 encoding exceeds `MDI_MAX_LINE_BYTES = 254` (256 minus line terminator minus one byte of slack).
- **`main.py send_cmd()`**: sends one write per line instead of the whole multi-line text in a single write; lines over the limit surface a red error row in the MDI console instead of being silently truncated by the machine. Existing `clear` semantics preserved.
- **`Controller`**: all `stream.send()` call sites (commands, realtime bytes, file commands) now serialize behind an `RLock`, so the status poll can no longer interleave with command writes.
- **`WIFIStream.send()`**: `socket.sendall()` instead of `socket.send()`.

## Testing

- `tests/unit/test_mdi.py`: 10 new tests — line splitting, CRLF, blank-line removal, exact-limit acceptance, UTF-8 byte-length measurement (multi-byte chars), mixed valid/rejected input.
- Full unit suite passes (255 tests; the two `hid`-dependent pendant test files were skipped locally on Windows because `hidapi.dll` is unavailable — unrelated to this change and covered by CI).
- `ruff check` / `ruff format` clean; `mypy --strict carveracontroller/machine/` clean; both import-linter contracts kept.
- App smoke-tested: boots cleanly with the changes.

## Out of scope

Pacing of rapid command bursts against the depth-4 makera command queue (and the resulting disconnects) is #710 territory — this PR only guarantees each MDI line arrives whole and individually.

### DIFF
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -104,6 +104,11 @@ class Controller:
         self.usb_stream = USBStream(log_sent_receive)
         self.wifi_stream = WIFIStream(log_sent_receive)
 
+        # Serializes writes to the active stream. The status poller (streamIO
+        # thread) and command senders (UI thread) share one socket/serial port;
+        # without a lock their writes can interleave mid-command.
+        self._send_lock = threading.RLock()
+
         # Reconnection properties
         self.reconnect_enabled = True
         self.reconnect_wait_time = 10
@@ -242,7 +247,8 @@ class Controller:
                     self._notify_usb_reset_blocked()
                     return
                 payload = line.encode() if isinstance(line, str) else line
-                self.stream.send(self.comms.encode_command(payload))
+                with self._send_lock:
+                    self.stream.send(self.comms.encode_command(payload))
                 if self.execCallback:
                     display = line if isinstance(line, str) else line.decode(errors="ignore")
                     # Strip ".lz" suffix for display
@@ -283,7 +289,8 @@ class Controller:
                 if isinstance(char, (bytes, bytearray)):
                     char = char[0]
                 payload.extend(self.comms.encode_realtime(int(char)))
-            self.stream.send(bytes(payload))
+            with self._send_lock:
+                self.stream.send(bytes(payload))
         except Exception:
             self.log.put((Controller.MSG_ERROR, str(sys.exc_info()[1])))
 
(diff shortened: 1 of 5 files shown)


SergeBakharev:
Same as https://github.com/Carvera-Community/Carvera_Controller/pull/727#issuecomment-5229083109

Please do minimal testing yourself before submitting a PR. Marking this as draft.

SergeBakharev:

PR #724, opened 2026-08-08
Pace queued command sends so bursts cannot stall the firmware link

### LINKED ISSUE #710: Features that send multiple commands in rapid succession cause Controller disconnect

oliveiracelso:
Fixes #710 on the controller side — a mitigation that keeps the link alive regardless of when Carvera-Community/Carvera_Community_Firmware#400 lands in firmware.

## Why bursts drop the connection

In Makera framed mode over WiFi the firmware has a **single pending-command slot**: `WifiProvider::on_idle` refuses to `receive_wifi_data()` while `makera_command_pending` is set, so from the moment a `CTRL_MULTI` frame is queued until `on_main_loop` dispatches it, the firmware reads **nothing** from the WiFi module — including `?` status queries. A burst of commands (resume-at-line sends ~10 back-to-back, CMM Workbench and probing sequences similar) keeps that slot occupied continuously; status responses stop, the heartbeat times out, and the controller declares the connection lost.

Over USB the framed command queue is 4 deep and `makera_cmd_queue_push` **silently drops** commands when it is full. Smoothie text mode assembles lines from a single `RingBuffer<char, 256>`, which a burst can also overrun.

## Approach: paced send queue with receive-based release

New `carveracontroller/machine/command_flow.py` (Kivy-free, `mypy --strict` clean): `CommandFlowControl`, a thread-safe FIFO where the next command is released only when:

- the machine has been heard from since the previous send (any complete parsed frame/line — evidence its receive path is live again), after a 20 ms minimum gap; **or**
- 0.5 s has elapsed with no response (fallback so an unresponsive machine cannot wedge the queue).

Wiring:

- `Controller.executeCommand()` enqueues by default; `streamIO` drains at most one command per loop iteration and calls `note_receive()` as it parses inbound frames. A single interactive command with a live connection still goes out on the next loop pass — only *bursts* get spaced.
- The queue is cleared on connect, disconnect, and abort.
- `immediate=True` bypasses the queue where waiting would be wrong:
  - `abort` (also drops everything still queued — aborting cancels the intent of pending commands) and `suspend`/`resume`;
  - `reset` flows (the connection closes right after; a queued reset would never send);
  - the **baud switch** flow, which sends `baud <n>` while `streamIO` is parked — with a queue and no drain it would deadlock.
- Realtime bytes (`?`, `!`, `~`, Ctrl-X/Y/Z) keep their existing direct path — they must never queue behind commands.

## Testing

- `tests/unit/test_command_flow.py`: 8 new tests — immediate first release, hold-until-heard, minimum gap, ack timeout, FIFO order, clear/reset semantics, pre-send receive not counting as an ack.
- Full unit suite passes (254 tests; the two `hid`-dependent pendant files skipped locally on Windows for lack of `hidapi.dll` — CI covers them).
- `ruff` clean, `mypy --strict carveracontroller/machine/` clean, both import-linter contracts kept.
- App smoke-tested: boots cleanly, status polling unaffected.

## Notes for review

- Worst-case pacing is one command per 0.5 s against a machine that answers nothing, and one per ~20 ms against a responsive one; resume-at-line's ~10 commands complete in well under 2 s even in the degraded case.
- `viewDiagnoseReport`'s periodic `diagnose` now also flows through the queue, which is intentional — it was equally capable of colliding with a burst.
- Mild overlap with #723 (both touch `executeCommand`); whichever lands second is a trivial rebase — happy to do it.

### DIFF
--- a/carveracontroller/Controller.py
+++ b/carveracontroller/Controller.py
@@ -20,6 +20,7 @@ from functools import partial
 
 from . import Utils
 from .CNC import CMDPAT, CNC, LASER_TOOL_NUMBER, PARENPAT, SEMIPAT, ZPROBE_TOOL_NUMBER
+from .machine.command_flow import CommandFlowControl
 from .protocols import MessageKind, ProtocolSession
 from .USBStream import USBStream
 from .WIFIStream import WIFIStream
@@ -103,6 +104,11 @@ class Controller:
     def __init__(self, cnc, callback, log_sent_receive=False):
         self.usb_stream = USBStream(log_sent_receive)
         self.wifi_stream = WIFIStream(log_sent_receive)
+
+        # Paces queued command sends so bursts (resume-at-line, CMM Workbench,
+        # probing sequences) cannot overrun the firmware's single pending-command
+        # slot / shallow command queue. Drained by streamIO.
+        self._cmd_flow = CommandFlowControl()
 
         # Reconnection properties
         self.reconnect_enabled = True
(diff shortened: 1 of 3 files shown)


Lyrkan:
> a thread-safe FIFO where the next command is released only when:
>  *  the machine has been heard from since the previous send (any complete parsed frame/line — evidence its receive path is live again), after a 20 ms minimum gap; or
>  *  0.5 s has elapsed with no response (fallback so an unresponsive machine cannot wedge the queue).

Unless I'm mistaken this wouldn't fix the issue.
The problem is mainly present with commands that take a long time to execute, for instance during probing.

Using the CMM example it would:
* send the probing command
* wait for one of those two conditions to be fulfilled
* buffer the following M118 => at that point wifi RX is blocked again until the probing command finishes processing
* detect a timeout and disconnect after 5s

SergeBakharev:
This is an interesting approach which might be a useful contribution irrespective of the outcome of https://github.com/Carvera-Community/Carvera_Community_Firmware/issues/400

However in addition to @Lyrkan comment that it is based in some flawed logic, I'm worried that it might be dangerous to allow extended queuing of commands. For example think of the case of jogging, or delayed spindle on. This could seriously hurt someone or their machines.

However I guess we are effectively doing this with the input buffers on the machine end, just with varying degrees of buffer space. Extending that buffer on the Controller side though might be were we should draw the line. Unless a Controller initiated workflow is expected to have extended (1s+) delays, I think it's safer to reject input that would be queued that far int the future

@oliveiracelso Do you have any thoughts on how to improve the safety aspect?

SergeBakharev:

PR #725, opened 2026-08-08
Stop the remaining-time countdown when the connection is gone

### LINKED ISSUE #712: Time continues if file playback aborted then disconnected from machine
Seems like a special variant of #604 that we didn't consider:

1. Start gcode playback
2. abort/stop
3. disconnect from the machine

After disconnecting the time remaining count down resumes

oliveiracelso:
Fixes #712.

## Root cause

Both disconnect paths (manual `close()` and the heartbeat-loss path) call `updateStatus()` right after closing the link — but the playing check there reads `CNC.vars["is_playing"]` / `CNC.vars["playedlines"]`, which hold the **last values the machine ever sent**. In the abort → disconnect repro, those values still reflect the aborted job, so the check lands in the `app.playing = True` branch and the one-second `_progress_smooth_clock` stays scheduled.

Meanwhile `_update_progress_smooth()` only freezes for `Hold/Pause/Wait/Tool`. During the abort the state passes through one of those (countdown frozen); after disconnecting the state becomes `N/A`, the guard no longer matches, and the countdown **resumes** against wall time — exactly the reported behaviour.

## Changes (surgical, 9 lines in `main.py`)

- `updateStatus()`: when the state shows no live machine (`N/A` or `Connected`), force the not-playing branch. That cancels the smooth progress clock, clears `app.playing`, and resets the progress bars. The interrupted-playback handling still runs first, so the resume-at-line position is recorded before the reset.
- `_update_progress_smooth()`: also freeze when the state shows no live machine — second layer of protection for any path where the clock survives a disconnect.

## Testing

- Unit suite passes (245 tests locally; `hid`-dependent pendant files skipped on Windows for lack of `hidapi.dll`, covered by CI). `ruff` clean.
- App smoke-tested: boots cleanly; file loading and playback simulation unaffected.
- The disconnect paths both funnel through `updateStatus()`, so the fix covers manual disconnect, heartbeat-loss disconnect, and the auto-reconnect flow.

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -5856,8 +5856,10 @@ class Makera(RelativeLayout):
             or (not app.selected_remote_filename and not app.selected_local_filename)
             or not self.selected_file_line_count
             or app.state in self._PROGRESS_TIMER_PAUSED_STATES
+            or app.state in (NOT_CONNECTED, CONNECTED)
         ):
-            # While held/paused, leave the last progress_info unchanged so both timers freeze.
+            # While held/paused (or with no live machine), leave the last
+            # progress_info unchanged so both timers freeze.
             return
         remaining_display = self._current_remaining_sec()
         filename = os.path.basename(app.selected_remote_filename or app.selected_local_filename)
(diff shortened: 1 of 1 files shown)


SergeBakharev:
Thank for this submission, I've changed the PR destination to the `release-2.2.0-RC2` branch. Please complete a CHANGELOG.md entry describing the fix.

SergeBakharev:

PR #726, opened 2026-08-08
Switch update checks to the GitHub releases API

### LINKED ISSUE #652: Switch version checking to use github releases API
Currently the version checking code reads the changelog.md from main branch to determine the latest version.

This puts in an awkward position when releasing RC releases or when we have merged but have build issues with a release.

Using the release API is a much better approach

oliveiracelso:
Fixes #652.

## What

Both update checks (controller and firmware) now use the GitHub releases API instead of scraping files from branch heads:

| Before | After |
|---|---|
| `raw.githubusercontent.com/...carvera_controller/main/CHANGELOG.md` + regex `[x.y.z]` | `api.github.com/repos/Carvera-Community/Carvera_Controller/releases/latest` → `tag_name` |
| `raw.githubusercontent.com/...carvera_community_firmware/master/version.txt` + regex | `api.github.com/repos/Carvera-Community/Carvera_Community_Firmware/releases/latest` → `tag_name` |

`/releases/latest` returns the newest **published, non-prerelease, non-draft** release — so merged-but-unreleased changes and RCs are never announced as updates, which is exactly the pain point in the issue.

## Changes

- `Utils.parse_github_release()`: tolerant `(version, notes)` extraction — accepts the dict that Kivy's `UrlRequest` produces for `application/json` responses as well as raw JSON strings/bytes, strips the `v` prefix from tags (`v2.1.0` → `2.1.0`, `v2.1.0c` → `2.1.0c`), and returns empty values for unusable payloads (404 bodies, missing/non-string tags) so callers treat them as "no release found".
- `check_for_updates()` sends a `User-Agent` header (required by the GitHub API) plus `Accept: application/vnd.github+json`.
- The upgrade popup now shows the **release notes** of the latest release instead of the raw changelog/version.txt dump, and version comparison uses the release tag directly instead of regexing the first `[x.y.z]` out of a markdown file.
- Existing update-flow structure kept intact: `fw_version_new`/`ctl_version_new`, `digitize_v` comparison, the idle-time `check_fw_version()` trigger and the badge properties all work as before.

## Testing

- 6 new unit tests for `parse_github_release` (dict/string/bytes payloads, `v`-prefix stripping, `name` fallback, null body, 404/garbage payloads); full suite passes (251 locally, `hid`-dependent pendant files skipped on Windows).
- Verified end-to-end against the live API: both endpoints fetched with the exact headers the app sends; controller parsed as `2.1.0`, firmware as `2.1.0c`, release notes extracted.
- App smoke-tested: boots cleanly, startup update check runs without errors.

## Notes

- Unauthenticated GitHub API allows 60 requests/hour/IP; the app makes 2 per launch, so rate limiting is not a concern.
- If a rate-limited/404 response ever arrives, the parser returns empty values and the app behaves exactly as it does today on a failed download (no update announced).

### DIFF
--- a/carveracontroller/Utils.py
+++ b/carveracontroller/Utils.py
@@ -8,6 +8,7 @@ __author__ = "Vasilis Vlachoudis"
 __email__ = "vvlachoudis@gmail.com"
 
 import hashlib
+import json
 import os
 import sys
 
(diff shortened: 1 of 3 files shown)


SergeBakharev:
The interactive testing here is very light. Have you run this yourself and made sure the firmware update screen correctly renders the release notes? The expectation for all PRs is that the submitter has done everything reasonable to review and test the PR on their side before marking the PR as ready for review (Open). This includes running the code against the own machine hardware.

Github release notes are rich text, not just ASCII like the current CHANGELOG.md I see no handling of that.

Marking this PR as draft. Please mark this as open when it's ready.

SergeBakharev:

PR #727, opened 2026-08-08
Show the WCS rotation angle on the coordinate-system button

### LINKED ISSUE #637: Rotation value not showing up in WCS tab

oliveiracelso:
Fixes #637, implementing the approach agreed in the issue discussion.

## What

When the active WCS has a rotation set, the coordinate-system button's minor text now **takes turns** between the WCS name and the rotation angle (`R: x.xxx°`), using the same `status_index` alternation as the feed, spindle and tool buttons. With no rotation, it shows the WCS name full-time, as before.

This addresses both sides of the discussion:

- the rotation angle is visible on the main screen again (the original report), and
- the WCS name is never hidden (the concern that led to #638 being closed — with a description set, replacing the minor text outright would have removed the name entirely).

The existing shrink-to-80% hint for a rotated WCS is kept, now driven by the same `has_rotation` condition.

## Also fixed

The guard flagged at the end of the issue: `populate_wcs_values()` checked `len(values) >= 5` but then unpacked **six** values, so a five-element entry would raise `ValueError` instead of being skipped. The guard now requires six values and unpacks `values[:6]` (B stays unused, per the discussion — the controller doesn't support 5-axis).

## Testing

- Full unit suite passes (245 locally; `hid`-dependent pendant files skipped on Windows), `ruff` clean.
- App smoke-tested: boots cleanly; coordinate-system button renders and alternates on the `switch_status` cadence.
- Not tested against a machine with a rotated WCS (no hardware here) — the alternation logic mirrors the surrounding feed/spindle/tool button code paths, which run on every status update.

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -1488,11 +1488,11 @@ class WCSSettingsPopup(ModalView):
         """Populate the WCS fields with parsed data from machine"""
 
         def update_ui(dt):
-            # wcs_data format: {'G54': [x, y, z, a, rotation], 'G55': [...], ...}
+            # wcs_data format: {'G54': [x, y, z, a, b, rotation], 'G55': [...], ...}
 
             for wcs, values in wcs_data.items():
-                if len(values) >= 5:  # Ensure we have X, Y, Z, A, rotation
-                    x, y, z, a, b, rotation = values
+                if len(values) >= 6:  # Ensure we have X, Y, Z, A, B, rotation
+                    x, y, z, a, b, rotation = values[:6]
 
                     # Store original values for comparison
                     self.original_values[wcs] = {"X": x, "Y": y, "Z": z, "A": a, "B": b, "R": rotation}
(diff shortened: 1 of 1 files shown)


SergeBakharev:
The interactive testing here is very light. The expectation for all PRs is that the submitter has done everything reasonable to review and test the PR on their side before marking the PR as ready for review (Open). This includes running the code against the own machine hardware.

If you don't have your own machine please use the simulator: https://github.com/f355/carvera_firmware_simulator

Marking this PR as draft. Please mark this as open when it's ready.

SergeBakharev:

PR #729, opened 2026-08-09
CMM Workbench: Detect unexpected machine states during probing

Lyrkan:
This PR adds a small check every time the CMM workbench spinner is refreshed (about 5 times/second when running) so that we can abort the current operation if the machine goes into a state other than "Run" or "Idle". The idea is to prevent users having to wait for the entire duration of the 30s timeout before resuming if an error occurs or if the controller gets disconnected from the machine.

**It should only be merged if we decide to abandon #713 since that one also contains a similar check**


Lyrkan:
@SergeBakharev Was this closed by mistake?

SergeBakharev:
@Lyrkan  Sorry about that, it deleted the branch release/2.2.x after I did a sync merge into develop. Probably closed these two PRs

Issue #509, opened 2026-02-09
Improve file management interface


acj:
We've had feedback that the file management workflow for transferring files to the machine and then loading them is buggy and confusing. I agree. Even if you understand that a gcode file needs to be on the machine for it to run, it's easy to lose track of which UI view you're looking at, and it's easy to click the wrong button and have to start again. 

I'd like to propose a two-column layout for the file browser that shows local files on the left and machine files on the right. Buttons in between the columns for transferring files in either direction. There are probably more modern UI examples, but WS_FTP has the right flavor:

<img width="638" height="330" alt="Image" src="https://github.com/user-attachments/assets/a49e9ad3-ee44-4d0d-abff-2e6ca873b7f7" />

Related ideas:
- If you select a local file and click the "Select" (or maybe "Load") button, we could automatically transfer it to the machine before loading it. The operator shouldn't need to worry about where the selected file is - if the machine needs it, then copy it to the machine.
- We could support copying entire directories. Might be useful for someone who moves between computers often.

[Discord thread](https://discord.com/channels/1341908765212934320/1470186041901056155/1470211246262583441)

Lyrkan:
I spent some time on this subject tonight and I'm not sure that a two-column layout is a good idea. It works if you're a bit used to that kind of UI but for newcomers it may end up being as confusing as the current layout. Also it may not be possible to scale it properly on small devices.


I put together some quick mockups of another proposal using two tabs that would make it easier to switch between the two sources:

<img width="1920" height="1055" alt="Image" src="https://github.com/user-attachments/assets/2faf1145-06fc-4a1a-869f-f914e87ebd65" />

<img width="1920" height="1055" alt="Image" src="https://github.com/user-attachments/assets/08561a50-954c-475e-9682-6826ee3088ac" />

<img width="1920" height="1055" alt="Image" src="https://github.com/user-attachments/assets/e314cd5d-f8d5-4bee-9b9f-9b0080e33a92" />

<img width="1920" height="1055" alt="Image" src="https://github.com/user-attachments/assets/c3c0ecb1-9a3a-47e1-8fab-1775f393d8da" />

<br/>
<br/>

The only thing that may be missing and that I don't know how to represent yet would be the destination path when uploading a local file (the idea being that it would use the active folder from the "Machine" tab).

Lyrkan:
[...]

PR #731, opened 2026-08-12
Add stock settings + simulation

Lyrkan:
Just a draft PR to show the current state of my experiments regarding stock settings / simulation in the controller using the recently added tool descriptions to carve voxels based on the tools' geometry:

Recording 2026-08-19 at 01 18 28

**Disclaimer so that no-one gets their hopes up too high:**
* Simulation is a slow. I'm not sure how much room for improvement there is without going the native route instead of just using Python (and that'd be quite the rabbit hole).
* Remeshing is even slower, which is the biggest issue. ~I had to disable the voxel rendering during playback because it's not even close to keeping up.~ EDIT: Actually with some more approximations (during playback, fixing themselves when pausing) and optimizations it is still lagging behind but is now able to catch up periodically so I've added this feature back behind a toggle in the settings.

---

**Stock settings:**

Stock settings (and enabling the simulation) can be access through a new icon in the G-Code viewer toolbar (since we now have a bit more space). I quickly made an icon for it but as usual: feel free to replace it if you've got a better idea.

By default the suggested stock size is calculated based on the content of the loaded file. It is using min/max X/Y coordinates with an offset based on the largest tool. For the Z axis, since it wouldn't really work due to retractions/clearance moves I reused the 95 percentile thing from the color scheme viewer "height" type. This is not perfect at all but good as a starting point.

In addition to that the code also tries to retrieve stock and origin information from the G-Code file headers.
For now I implemented Makera Studio (from my understanding of what they currently put in the headers - since there is no spec and that Makera Studio is missing a lot of things I was not able to test all cases) and Fusion 360 (with an updated Carvera.cps - I'll open a PR for it quite soon). A new setting, enabled by default, allows to directly show the stock (without starting the simulation) if we were able to get those info from the headers.

---

**How the simulation works (sorry for the wall of text):**

When the stock is set up in the controller (either automatically or manually) the user can choose between multiple kinds of "carver". Each one of them has a different approach and differents pros/cons (how fast/memory intensive they are, whether or not they support undercuts, etc.). In each case there is also a quality settings that determines the amount of subdivisions with values based on the largest side of the stock.
<br/>

**Heightmap carver**
This is a "dumb" carver that, as its name implies is based on a heightmap of the stock (2D grid with the current height of each column). It only works for 3 axes jobs and doesn't support undercuts. However it's the fastest option since we only have to change the height of each column during simulation.

I've set it as the default choice for 3 axes jobs that don't have any undercut (based on the absence of any tool that could cause one, ie. lollipop endmills, thread mills, etc.).
<br/>

**Cylindrical carver**
Available for 4 axes jobs, it uses dexels which work kind of the same as heightmaps but with heights that are based on a location and angle on a cylinder (so the "heights" are coming from the center of the cylinder). Due to that:
* it can't represent undercuts either
* it also cannot represents move that would carve at negative Z
* the precision changes depending on the current height (more dense towards the center)

Set as the default choice for 4 axes jobs that don't have any undercut.
<br/>

**Voxel carver**
Here the stock is represented as a 3D grid of voxels. Basically it takes the longest axis of the stock and chooses the amount of voxels based on the selected level target (clamped so that voxel size end up between 0.1mm and 5mm), for instance (values may change later on):
* for a 200x20x10mm stock and a low quality setting (100 voxels), the grid will be composed of 100x10x5 voxels (ie. 2mm voxels)
* for a 50x100x5mm stock and a high quality setting (500 voxels), the grid will be composed of 250x500x25 voxels (ie. 0.2mm voxels)

<br/>

Instead of directly storing all the voxels, the grid is populated with 16x16x16 voxels clusters.
Each clusters has 3 states:
* Full (doesn't consume any memory): all voxels in the cluster are solid. For rectangular stock this is the default state for all clusters. For cylinders it's a bit more complex (and probably not that optimal): they are seeded first so voxels outside the cylinder are set to empty.
* Empty: all the voxels have been carved out (or never existed in the case of a cylinder)
* Partial: at least one voxel in the cluster is missing but some are still solid.

image

<br/><br/>

This allows to reduce the memory footprint and also optimize carving since we can:
* directly ignore clusters that are empty
* avoid comparing every single voxels in a cluster to the tool if it is still full and not intersecting

<br/>

The carving process itself is quite straightforward (if you ignore optimizations): take the tool profile (excluding the shank) and check if it intersects a voxel centre during its travel. If so the voxel is removed.

<br/>

Some implemented optimizations:
* already carved clusters/voxels are directly ignored
* fast move are ignored (even though they could in theory have an impact it should probably not be the case)
* before checking a cluster/voxel we check if the tool path is actually close enough using its XY coordinates and an increased bounding box
* depending on the tool shape some checks are simplified (for instance flat endmills can skip profile interpolation since they have a constant radius)
* instead of processing every single move we merge almost collinear same-tool cuts into a single straight line (to a certain extent: at most 64 segments, capped at ~8mm or 16 voxels (whichever is larger), 0.25 voxel deviation, doesn't go past next checkpoint (see below))

<br/>
<br/>

Then, for all those carvers, comes the issue of being able to move back and forth smoothly in the playback timeline.
Reconstructing everything is obviously not an option since it would take forever, so instead I added a checkpoint system which works as follows:
- when setting up the simulation an option allows to specify the amount of checkpoints to store (currently: 256, 1024 or 4096, but I'll probably tweak them later on)
- those checkpoints are spread equally based on path vertex index
- when carving if we go past the location of the next checkpoint we store the current state of the grid (non-solid chunks only)
- when going back in time we restore the grid to the state of the preceding checkpoint (last full snapshot and at most 3 deltas, see below) and only have to re-carve the short remaining gap (jumping forward can also use checkpoints the same way).

But, in order to be a bit more memory-efficient, instead of directly storying the full grid state everytime we only take a full snapshot every 4 checkpoints (this also may be subject to change).
A full snapshot is still only non-solid cluster whereas the intermediate checkpoints only store the clusters that changed since the previous checkpoint.

<br/>

Carving happens in a worker thread and depends on the current playback state:
* when paused we try to process the file ahead of the current position (using a second sparse grid) to generate checkpoints without moving the live occupancy
* when playing we try to carve occupancy up to the current position so it is synced with the tool (we can then patch the carved mesh if the live preview is enabled in the settings)

---

**Some screenshots:**

**Stock/simulation settings:**
stock-simulator-02

<br/><br/>

**Disabled simulation (stock only):**

image

<br/><br/>

**Enabled simulation:**

image

stock-simulator-05

<br/><br/>

**Cylindrical stock:**

image

<br/><br/>

**4th axis job:**

stock-simulator-11

<br/><br/>

**Jobs with laser paths:**

image

image

Recording 2026-08-19 at 19 00 29


SergeBakharev:
Looking good!

SergeBakharev:
This is quite exciting! I was sure that this wasn't feasible, first because of stock dimensions not in metadata and second because of processing time / GIL limitations of python. But you have cut through all the problems! Very impressive!

Lyrkan:
> first because of stock dimensions not in metadata 

That's still something I have to look at, for now the dimensions are not set automatically but:
* I know Makera Studio adds some info regarding the stock, and it would not be that hard to parse it
* Maybe we could do the same with other post processor but I haven't worked on that yet
* As a fallback I was thinking about taking the `(max coords - min coords)` of all paths excl. fast moves + the max tool diameter as an offset **(edit: implemented that with a twist for the Z axis, see the original post for details)**

Lyrkan:
Added 4th axis support, took quite a long time to generate the full Nefertiti example but at least it is recognizable:

image
<br/><br/>
The hole on the left side is unrelated I think, I simply concatenated the rough and finishing passes + removed a few instructions and for some reason it adds a path there after switching to T2...:

```
G01X63.24Z8.00A-79985.29
G01X63.64Z8.00A-79972.94
G00X63.64Z27.00A-79972.94
G92.4A0S0
T2M6
M3 S12000
G01Z30.00 F1500
G01Y0.00Z30.00
```

image

Lyrkan:
I'll write more about it when updating the original post later but this weekend I worked on two new carving strategies in addition to the voxels one.

Since most jobs probably don't have undercuts (or they are not that important to view) we can improve performances by just representing the stock as a heightmap, or for 4th axis job a dexels grid (which is kind of the same thing but wrapped around a cylinder).

This allows for faster processing (**but again: don't hope for a miracle, we're still limited by Python**) even at higher res.

The carver can be chosen from the stock settings popup and defaults to either voxel or heightmap/dexel depending on whether or not the current file contains a tool that can cause undercuts.


image

image

Lyrkan:
Added support for jobs that have laser paths.

It basically works by generating a 2D texture that acts as a decal on the remaining stock  (more pictures in the MR description):

image

Lyrkan:
Fixed some things which improved the accuracy of the simulation, in particular for 4th axes jobs where some paths were incorrectly detected as collinear (resulting in the ugly stripes visible in my previous Nefertiti screenshot).

Also took the time to implement a basic material selector:

stock-simulator-11

stock-simulator-12

SergeBakharev:
Looking great, love your work!

Lyrkan:
Removing the draft flag!
Definitely far from perfect but I think it should be good enough for a first version.

<br/>

Small note though: I had to bump the Android min API/NDK version to 24 (instead of 21 I think) in order to add Numpy. I don't think this should be a big issue since it is only relevant for devices with a version of Android < 7 (not updated since 2015).
This change also required to pin `charset-normalizer` to v2.1.1 because newer versions (automatically retrieved after the Android version bump) [caused the build to fail](https://github.com/kivy/python-for-android/issues/2755) with our version of P4A.

<br/>

I've also created two post-processor PRs ([FreeCAD](https://github.com/Carvera-Community/Carvera_Community_Profiles/pull/53) and [Fusion](https://github.com/Carvera-Community/Carvera_Community_Profiles/pull/51)) so that stock settings and origins can automatically be exported from CAM.

SergeBakharev:
I don't remember the check boxes in the Facing tool being so hard to see (I'm now noticing that have the same problem), but on my machine it's nearly impossible to see the boxes! 

Screenshot 2026-08-28 at 11 11 03 am

Lyrkan:
@SergeBakharev Can you check if the last commit fixes it?

Older popups used a `MCQCheckBox` with a forced white `color`. I didn't directly used that because from what I understand the point was to have a `MCQCheckBox` followed by a `MCQLabel` (and in those popup labels are before the checkbox) but instead moved the `color` declaration to all instances of `Checkbox` (this popup + the facing tool).

Also tests are failing due to (I think) a timing issue, I'll try to fix that as well.

SergeBakharev:
Looks good now!

Screenshot 2026-08-28 at 2 43 28 pm

SergeBakharev:
I love this, agreed that this isn't perfect but I'm inclined to merge it without much change and tweak after more testing and feedback from others.

Even the default option showing the stock wireframe is a very welcome addition. The carver automatic selection is a great, though for the sorts of things I make there wasn't a huge difference in performance between Voxel and Heightmap/Cylindrical.

I'm not sure if the logic to select Cylindrical for 4th axis can be based on the absence of undercutting tools since you can rotate the work into a position that is "undercutting". Maybe 4th axis programs should default to Voxel. Here is an example where it defaulted to cylindrical and looked strange. Also not sure what happen with the part floating in air on the Voxel sim. Here is a comparison of Voxel vs Cylindrical:
[4th_undercut_test.cnc.txt](https://github.com/user-attachments/files/31608895/4th_undercut_test.cnc.txt)

Screenshot 2026-08-30 at 4 16 45 pm

Screenshot 2026-08-30 at 4 15 42 pm

I also some how managed to get the tool to desync from the replay in a different file. But I wasn't able to reproduce the issue again.
Screenshot 2026-08-30 at 3 43 31 pm.

A few related "follow up" additions for separate PRs perhaps:
* Use the stock shape for margin drawing
* Controller setting to default the cutting simulation to enable/disable if stock data is present in gcode
* Middle mouse click to move the camera
* Using the Controller offline is a bit clunky to view files, you have to go into "upload" then you can select a local file for viewing. I think a move to #509 would help address this, or at the very least when offline it should default to local view...

Lyrkan:
> I'm not sure if the logic to select Cylindrical for 4th axis can be based on the absence of undercutting tools since you can rotate the work into a position that is "undercutting". Maybe 4th axis programs should default to Voxel. 

I think it could be improved by checking if the job contains off-axis Y moves, let me try implementing that.

> Also not sure what happen with the part floating in air on the Voxel sim. 

I looked at your file and it is because the endmill is not long enough so it never intersect with those voxels.

image

> I also some how managed to get the tool to desync from the replay in a different file. But I wasn't able to reproduce the issue again

Hm, when simulating during playback it can happen since it's always a race between the two, but when paused it is supposed to catch up from the nearest previous checkpoint. I'll investigate.

> * Use the stock shape for margin drawing

Not entirely sure what you mean by that one. Would it be drawing the outline of the final state on the stock?

> * Controller setting to default the cutting simulation to enable/disable if stock data is present in gcode

I actually didn't add a setting for that on purpose since some jobs can be quite heavy to simulate and wanted to kind of force people to go through the simulation settings. Maybe once it's optimized enough and we've got enough feedback regarding the default settings?

> * Middle mouse click to move the camera

Moving the camera is already mapped to right click. Maybe we should add a way to customize that, a bit like FreeCAD does?

image

<br/>
<br/>

> * Using the Controller offline is a bit clunky to view files, you have to go into "upload" then you can select a local file for viewing. I think a move to [Improve file management interface #509](https://github.com/Carvera-Community/Carvera_Controller/issues/509) would help address this, or at the very least when offline it should default to local view...

Yeah, even knowing how it works I sometimes forget whether I am looking at the machine files or the local ones...

Lyrkan:
The cylindrical/voxel default for 4th axis jobs should be better now.

When parsing the file it now count the number of feed moves that have an |Y| of more than 1mm, and if they represent more than 5% of the total feed moves the 4th axis carver defaults to voxels.

In practice:
- something like your file => voxels
- something like Nefertiti => cylindrical

SergeBakharev:
Nice, I think the cylindrical/voxel logic for 4th axis jobs is sound.

> > Use the stock shape for margin drawing
>
> Not entirely sure what you mean by that one. Would it be drawing the outline of the final state on the stock?

I meant the initial shape of the stock, I don't think the final shape is all that useful to draw.

Issue #395, opened 2025-10-01
Probing popup accessiable when playback is suspended


SergeBakharev:
We should disable the probing UI button when playback is suspended.

This is what the user sees:

<img width="4000" height="2252" alt="Image" src="https://github.com/user-attachments/assets/0630280d-ffea-49a7-bb0d-0d520540365b" />

PR #732, opened 2026-08-12
Fix config load fail on QuickLZ-compressed downloads (ftype=lz)

ExaltedRaddix:
- Detect QuickLZ signature 0x0000 at start of cached config.txt
- Call decompress_file() BEFORE parsing (same path as gcode files use)
- Use open(encoding='utf-8', errors='replace') instead of Windows default charmap locale, fixing the 'charmap codec ... position 51' loop on non-English Windows installations (Community Firmware 2.2.0+)
- Fixes the infinite reload loop when controller connects via Makera framed protocol and firmware sends ftype=lz XMODEM/FILE transfers

### DIFF
--- a/carveracontroller/main.py
+++ b/carveracontroller/main.py
@@ -4793,7 +4793,36 @@ class Makera(RelativeLayout):
                 config_path = os.path.join(self.temp_dir, "config.txt")
                 if not os.path.exists(config_path):
                     raise FileNotFoundError(f"Cached config not found: {config_path}")
-                with open(config_path) as f:
+
+                # BUGFIX #1: Check for QuickLZ compression (first two bytes 0x00 0x00)
+                # and decompress if needed. The firmware compresses files for XMODEM
+                # transfer (ftype = lz), but config downloads didn't go through the
+                # normal decompress path, causing 'charmap codec' decode errors.
+                with open(config_path, "rb") as f_raw:
+                    first_bytes = f_raw.read(2)
+                if first_bytes == b"\x00\x00":
+                    logger.info("Config file is QuickLZ-compressed. Decompressing.")
+                    decompressed_path = config_path + ".decompressed"
+                    if self.decompress_file(config_path, decompressed_path):
+                        try:
+                            os.replace(decompressed_path, config_path)
+                        except OSError:
+                            shutil.copyfile(decompressed_path, config_path)
+                            try:
+                                os.remove(decompressed_path)
+                            except OSError:
+                                pass
+                    else:
+                        logger.error("Failed to decompress config file. Trying raw read anyway.")
+                        try:
+                            os.remove(decompressed_path)
+                        except OSError:
+                            pass
+
+                # BUGFIX #2: Always read as UTF-8 with errors='replace' to avoid
+                # Windows 'charmap' codec errors when the default locale charmap
+                # does not support every byte.
+                with open(config_path, encoding="utf-8", errors="replace") as f:
                     config_string = "[dummy_section]\n" + f.read()
                 # remove notes
                 config_string = re.sub(r"#.*", "", config_string)


SergeBakharev:
Thank you for this submission, but at this point I'm not inclined to accept this PR as it addresses a very niche issue and has some assumptions that I don't think are solid. But let's have a discussion, maybe my understanding is incorrect.

The scenario this PR addresses is specifically what results after uploading a config.txt sourced from the [Makera](https://github.com/MakeraInc/CarveraFirmware/blob/master/src/config.default) or previous versions of our firmware repo using the Controller.

I believe this situation is the case for a number or reasons:
* The charmap issue results from non-ascii characters being in this default config files.[ I address this in our repo last week.](https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/395). Real machines don't have this issue because the config.txt that is shipped from Makera doesn't have those characters.
* Actual configuration changes made by the Controller are made via `config-set` command, not by uploading a config.txt. For this reason there wouldn't be a .lz file to download, and even if there was it would be out of date from the live config.txt file. 

> open(encoding='utf-8', errors='replace') instead of Windows default charmap locale

I don't think this is appropriate because the machine itself doesn't support anything other than ASCII. Why should the Controller support an encoding that the machine doesn't? If there is non-ascii content something unexpected has happened. I suppose we could open it with `encoding='ascii', errors='replace'` to guard against this specific situation. Anything other than ASCII can't actually be `config-set` though, so this is more of a defensive measure.

I see the code here tries to pickup a QuickLZ signature and run a decompress before parse, and the "signature" it is using is what is currently used by gcode file downloads by the controller. However QuickLZ doesn't actually have a signature or magic number/header, and that's just some dodgy code we inherited. I'm not sure we should trust this way of identifying if a file is actually lz compressed or not. It's just looking for the high 16 bits of the first big-endian block length. A valid uncompressed text file starting with two NUL characters would match this check, unusual but still simple to do. On the flip side valid LZ compressed files (made with external tools) may not start with the same 16 bits.

So I think the correct course of action here is:
1. Check the firmware behavior, what happens when a config.txt that was uploaded via controller (thus has .lz and .md5 side files), and then subsequently updated via `config-set` or `config-delete`? It should not send the .lz file it previously received, if it does it's a bug and should be fixed in the firmware
2. We should have a think about what the actual desired behavior should be, given that I don't think there is a reliable way of telling if a file is LZ compressed or not. For now it might be just most prudent to accept this `0x00 0x00` signature. If so it shouldn't be done in `finishLoadConfig()` but in the comms code.
3. How ever you managed to upload the config.txt should be revised to not send a compressed file, I don't think the benefit is worth it. Especially when it comes to supporting machines running the Makera firmware that will likely still be affected by 1.

I personally think this is another problem related to the awkward implementation of LZ in the codebase, and maybe we should have a discussion about ripping it out altogether for transport-level compression. Or at the very least a compression type that has a defined signature. Or if we are all cool with two NUL characters meaning LZ compressed...

ExaltedRaddix:
Hi Sergey,
Thanks for the detailed reply! Quick clarification first, because this is a key misunderstanding:
I never uploaded a config.txt at all. I only downloaded the existing config from the machine via the Makera framed protocol (WiFi, port 2222) to debug why the Controller entered an infinite loop on my German Windows machine after upgrading the machine to Community Firmware 2.2.0-RC1. The Controller is what triggered the loop immediately on first connect.
So the .lz side file that the Controller receives is not something I created via upload — it is what the firmware itself produces/stores when serving config.txt via the Makera protocol file transfer . That means this is not a niche "user uploaded an old config" scenario at all — it happens on a completely stock machine (no CLI uploads ever done, no manual config tampering) simply by installing 2.2.0-RC1 and connecting with the Controller on non-English Windows.
That said, I fully agree with most of your technical points:
1. ASCII + errors='replace' instead of utf-8 is the correct choice — you're right, the machine only speaks ASCII. My utf-8 pick was defensive (I saw non-ASCII in the file I got, but that non-ASCII itself was probably the result of the bug). If there's non-ASCII in the config at all, defensive replacement is the right behaviour, not full encoding support. I can adjust that.2. 0x00 0x00 isn't a real QuickLZ signature — correct, and you're right that doing this detection in finishLoadConfig() is too late/wrong layer. Moving the decompression into the comms code (where the file transfer ftype=lz header is actually received and the file is written to disk) is the correct place. The framing/protocol code knows whether it received a compressed transfer or not, because it has the ftype field — it should decompress before writing the cached file, so the consumer never sees .lz bytes on disk at all. That eliminates the "0x0000 magic" heuristic entirely, which is definitely fragile.3. Stale .lz vs live config-set state — point taken, that's definitely a firmware question. In my case though, no .lz was ever uploaded from my side. I would value your input: why does the firmware cache and serve .lz files even for configs that were never uploaded as such? If the firmware always transparently compresses incoming/outgoing config.txt transfers for bandwidth, that's a protocol feature the Controller must handle — and the Controller must handle it correctly in the comms layer, not guess after the fact.On ripping out QuickLZ for transport-level compression entirely: I'd support that discussion. A format with a real header/magic would avoid this whole class of bugs.
If you agree with the direction above I can rework PR #732 to:
- Move the decompression logic out of main.py entirely- Decompress in the Makera protocol code immediately upon receipt (when the frame ftype explicitly says lz — no magic byte heuristic needed at all, since we have explicit metadata from the protocol)- Change the file open in finishLoadConfig() to encoding='ascii', errors='replace' as a purely defensive measureThis would address every concern you raised while still fixing the actual infinite loop that happens on stock 2.2.0-RC1 + non-English Windows. Let me know if that's the way to go and I'll update the PR.
Cheers!



Gesendet: Donnerstag, 13. August 2026 um 09:33
Von: SergeBakharev ***@***.***>
An: Carvera-Community/Carvera_Controller ***@***.***>
CC: ExaltedRaddix ***@***.***>,Author ***@***.***>
Betreff: Re: [Carvera-Community/Carvera_Controller] Fix config load fail on QuickLZ-compressed downloads (ftype=lz) (PR #732)

SergeBakharev left a comment (Carvera-Community/Carvera_Controller#732)
Thank you for this submission, but at this point I'm not inclined to accept this PR as it addresses a very niche issue and has some assumptions that I don't think are solid. But let's have a discussion, maybe my understanding is incorrect.
The scenario this PR addresses is specifically what results after uploading a config.txt sourced from the Makera or previous versions of our firmware repo using the Controller.
I believe this situation is the case for a number or reasons:

The charmap issue results from non-ascii characters being in this default config files. I address this in our repo last week.. Real machines don't have this issue because the config.txt that is shipped from Makera doesn't have those characters.
Actual configuration changes made by the Controller are made via config-set command, not by uploading a config.txt. For this reason there wouldn't be a .lz file to download, and even if there was it would be out of date from the live config.txt file.


open(encoding='utf-8', errors='replace') instead of Windows default charmap locale I don't think this is appropriate because the machine itself doesn't support anything other than ASCII. Why should the Controller support an encoding that the machine doesn't? If there is non-ascii content something unexpected has happened. I suppose we could open it with encoding='ascii', errors='replace' to guard against this specific situation. Anything other than ASCII can't actually be config-set though, so this is more of a defensive measure.

I see the code here tries to pickup a QuickLZ signature and run a decompress before parse, and the "signature" it is using is what is currently used by gcode file downloads by the controller. However QuickLZ doesn't actually have a signature or magic number/header, and that's just some dodgy code we inherited. I'm not sure we should trust this way of identifying if a file is actually lz compressed or not. It's just looking for the high 16 bits of the first big-endian block length. A valid uncompressed text file starting with two NUL characters would match this check, unusual but still simple to do. On the flip side valid LZ compressed files (made with external tools) may not start with the same 16 bits.
So I think the correct course of action here is:

Check the firmware behavior, what happens when a config.txt that was uploaded via controller (thus has .lz and .md5 side files), and then subsequently updated via config-set or config-delete? It should not send the .lz file it previously received, if it does it's a bug and should be fixed in the firmware
We should have a think about what the actual desired behavior should be, given that I don't think there is a reliable way of telling if a file is LZ compressed or not. For now it might be just most prudent to accept this 0x00 0x00 signature. If so it shouldn't be done in finishLoadConfig() but in the comms code.
How ever you managed to upload the config.txt should be revised to not send a compressed file, I don't think the benefit is worth it. Especially when it comes to supporting machines running the Makera firmware that will likely still be affected by 1.

I personally think this is another problem related to the awkward implementation of LZ in the codebase, and maybe we should have a discussion about ripping it out altogether for transport-level compression. Or at the very least a compression type that has a defined signature. Or if we are all cool with two NUL characters meaning

SergeBakharev:
@ExaltedRaddix 

Something still isn't adding up here. We have had Makera framed protocol in the Controller/Firmware for nearly a month. There are over 50 downloads of the 2.2.0-RC1, and a half dozen contributors running the bleeding edge dev version. This is the first report of this issue. If it occurred as described everyone (including me) would be experiencing it, but we are not.

Can you share your config.txt? Does it have the issue resolved in https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/395 ? If it does then this is not the original config.txt that came with the machine, and someone at some point has transferred the file from the git repo.

> why does the firmware cache and serve .lz files even for configs that were never uploaded as such?

I haven't looked deeply but I don't think there is special handling for the config.txt, it's just following the same code path as everything else. If a .lz exists in it's side storage, it will use it. We didn't develop this feature, so I would be guessing but I think it makes sense since the processing power on the machine is very limited to save cycles on compression it keep both the compressed and decompressed version. I think simple logic could be added to md5 check the file, and if the stored md5 (which came at the same time as the .lz) doesn't match the firmware should not use the .lz file

> Moving the decompression into the comms code (where the file transfer ftype=lz header is actually received and the file is written to disk) is the correct place. The framing/protocol code knows whether it received a compressed transfer or not, because it has the ftype field — it should decompress before writing the cached file, so the consumer never sees .lz bytes on disk at all. That eliminates the "0x0000 magic" heuristic entirely, which is definitely fragile

Unfortunately this is not sufficient. ftype=lz does not guarantee that the file transferred is LZ compressed. The firmware has a fallback to use the uncompressed version if no .lz file exists. This is the case for all the example files, and will be the case for any files transferred out-of-bound. `ftype=lz` is more like a signal that the machine supports lz and the transfer should be preferred to be compressed. But it's not a guarantee. It would be curious to see how the firmware handles detection of compressed vs uncompressed, what "signature" does it look for?

ExaltedRaddix:
Hi Sergey,
Thanks a lot for the quick reply and the detailed thoughts! First, a quick correction up front: I never uploaded any configs to the machine at all. I only used carvera-cli config get / config get-all to download the configs from the stock v2.2.0-RC1 firmware to check whether non-ASCII characters were causing the issue on my end. So this bug reproduces 100% on stock firmware + stock Carvera Community Controller connecting from a German Windows machine, with zero modifications from my side.
To your points:
1. Why only you? — The crash is OS-locale dependent. The code calls Python open() without an explicit encoding= parameter, which falls back to the system's default locale encoding. On German (and many non-English) Windows installs that is cp1252 / charmap , while on most English developer machines it ends up being UTF-8. When the firmware delivers a QuickLZ-compressed .lz file and the controller tries to read those binary bytes through charmap as text, you reliably get charmap codec can't decode byte 0x… in position …: character maps to <undefined> . On UTF-8 systems this crash is intermittent (depends on whether the LZ byte-stream happens to form valid UTF-8 sequences), which would explain why it only surfaces for some users.2. PR #395 Fix non-ASCII in config — Read through it, thanks for cleaning up the secÂ² mojibakes in the default configs! To verify this against my own machine I ran a byte-level scan ( bytes[i] > 127 ) across all config dumps I pulled from my Carvera today : config.txt , config-get-all.txt , config-get-all.keyvalue.txt , my previous config_aus_neuer_firmware.txt dump, and the fertig/ pair. Result: zero non-ASCII bytes anywhere in any of them. So the hypothesis "you had old repo-default configs with non-ASCII on your machine" cannot be confirmed from the actual data on my machine. The crash happens even with perfectly clean 7-bit ASCII content, purely because the .lz binary stream gets interpreted as text without prior decompression.3. Why .lz for configs / how to detect LZ? — I pulled the full Player.cpp from the Community Firmware Dev branch and read through the relevant sections to see exactly what heuristics the firmware itself uses:      - Format: No global file magic / header at all. The firmware does block-wise decompression: 4 bytes Big-Endian u32BlockSize per block, followed by that many QuickLZ-compressed bytes, and at the very end of all blocks a 2-byte Big-Endian CRC16 checksum.   - For small files like config.txt (< 64 KB compressed) the first block size is always ≤ 0xFFFF . Since the size is stored Big-Endian (high bytes first), the first two bytes of any small .lz file are automatically 0x00 0x00 . That is exactly the simple heuristic I used in the first version of the fix.   - The firmware's own download logic: fwfs::fopen("/.lz/filename") first, only if that returns NULL does it fall back to the uncompressed original. So exactly as you said: ftype=lz in the Makera frame means "peer offers LZ transfer if the .lz file exists" — it is not a hard guarantee, the fallback always has to work.Proposal for the final fix (I can update PR #732 to whichever you prefer):
- Keep the 0x00 0x00 prefix heuristic exactly as-is, because it mirrors the firmware's own behaviour and is reliable for any text-based file smaller than 64 KB (configs, G-code preambles, etc.).- Replace the strict UTF-8 decoding with errors='replace' so any future odd non-ASCII bytes just get substituted instead of crashing the load.- If you would rather have it architecturally cleaner: move the whole LZ detection + decompression logic into the comms layer ( protocols/makera.py ) instead of doing file-level fixups in main.py . Happy to refactor PR #732 that way.Let me know which direction you want to go and I will force-push an updated PR #732.
Thanks again,Johannes (ExaltedRaddix)



Gesendet: Freitag, 14. August 2026 um 00:54
Von: SergeBakharev ***@***.***>
An: Carvera-Community/Carvera_Controller ***@***.***>
CC: ExaltedRaddix ***@***.***>,Mention ***@***.***>
Betreff: Re: [Carvera-Community/Carvera_Controller] Fix config load fail on QuickLZ-compressed downloads (ftype=lz) (PR #732)

SergeBakharev left a comment (Carvera-Community/Carvera_Controller#732)
@ExaltedRaddix
Something still isn't adding up here. We have had Makera framed protocol in the Controller/Firmware for nearly a month. There are over 50 downloads of the 2.2.0-RC1, and a half dozen contributors running the bleeding edge dev version. This is the first report of this issue. If it occurred as described everyone (including me) would be experiencing it, but we are not.
Can you share your config.txt? Does it have the issue resolved in Carvera-Community/Carvera_Community_Firmware#395 ? If it does then this is not the original config.txt that came with the machine, and someone at some point has transferred the file from the git repo.

why does the firmware cache and serve .lz files even for configs that were never uploaded as such?

I haven't looked deeply but I don't think there is special handling for the config.txt, it's just following the same code path as everything else. If a .lz exists in it's side storage, it will use it. We didn't develop this feature, so I would be guessing but I think it makes sense since the processing power on the machine is very limited to save cycles on compression it keep both the compressed and decompressed version.

Moving the decompression into the comms code (where the file transfer ftype=lz header is actually received and the file is written to disk) is the correct place. The framing/protocol code knows whether it received a compressed transfer or not, because it has the ftype field — it should decompress before writing the cached file, so the consumer never sees .lz bytes on disk at all. That eliminates the "0x0000 magic" heuristic entirely, which is definitely fragile

Unfortunately this is not sufficient. ftype=lz does not guarantee that the file transferred is LZ compressed. The firmware has a fallback to use the uncompressed version if no .lz file exists. This is the case for all the example files, and will be the case for any files transferred out-of-bound. ftype=lz is more like a signal that the machine supports lz and the transfer should be proffered compressed. But it's not a guarantee. It would be curious to see how the firmware handles detection of compressed vs uncompressed, what "signature" does it look for?

SergeBakharev:
Johannes,

The crash cannot be OS-Locale dependent. The reason I'm so confident that it's not because of your German locale is because Python's default filesystem encoding is always `cp1252` on Windows irrespective of locale, per the Python documentation:

> [open()](https://docs.python.org/3/library/functions.html#open)
>  In text mode, if encoding is not specified the encoding used is platform-dependent: [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) is called to get the current locale encoding.

> [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding)
> ...
> On Windows, return the ANSI code page.

Please have a think about you might have done that is special. If you pull the SD card out of the machine and examine the creation date of the config file there might be a clue.

I looked at the player.cpp firmware code myself. [It determines if a file is LZ compressed](https://github.com/Carvera-Community/Carvera_Community_Firmware/blob/bc4d29921214d758fc5ff5f0f3a364e4856706f8/src/modules/utils/player/Player.cpp#L1662) in an upload by the file extension suffix `.lz`:
``` cpp
//if file is lzCompress file,then need to Decompress
start_pos = filename.find(".lz");
string srcfilename=lzfilename;
string desfilename= filename;
if (start_pos != string::npos) {
	desfilename=filename.substr(0, start_pos);
	if(!decompress(srcfilename,desfilename,u32filesize,stream))
       {
           if (communication_protocol == PROTOCOL_MAKERA) { 
               sprintf(error_msg, "error: error in decompressing file!\r\n");	
           }
		goto upload_error;
       }
   }
```

We can't do the same in the Controller, since we are requesting the file /sd/config.txt, and per your claim receiving a .lz compressed stream instead.

Johannes, I've looked further at the firmware code and in fact there IS guardrails to prevent a compressed /sd/config.txt from being returned. [Player::download_command](https://github.com/Carvera-Community/Carvera_Community_Firmware/blob/bc4d29921214d758fc5ff5f0f3a364e4856706f8/src/modules/utils/player/Player.cpp#L2093) always maps through [change_to_lz_path()](https://github.com/Carvera-Community/Carvera_Community_Firmware/blob/bc4d29921214d758fc5ff5f0f3a364e4856706f8/src/libs/utils.cpp#L313), which only knows about /sd/gcodes/.lz/:

``` cpp
// Change from origin path to quicklz file sub path
std::string change_to_lz_path( std::string origin )
{
	unsigned found = origin.find("gcodes/");
	string filename = origin.substr(found + 7);	
	string path = "/sd/gcodes/.lz";
	DIR * d = fwfs::opendir(path.c_str());
	if(NULL == d )
	{
		fwfs::mkdir(path.c_str(), 0);
	}
	else
	{
		closedir(d);	
	}
	return "/sd/gcodes/.lz/" + filename;
}
```

As an aside, I can't help but notice that your correspondence is written with AI assistance. While I have no issue with it's use for translation purposes, I can't help but feel that I've been put on a wild goose chase for a bug that doesn't exist by a confidently-wrong chatbot.

I won't take this PR any further until you can show the code path that would lead to a LZ compressed config.txt being returned in response to the command `download /sd/config.txt`.

The only reasonable explanation, like I've said from the start, is that you've done something strange on the Control side. Maybe uploaded a double LZ compressed file, or something else. Are you developing your own Controller modifications?

ExaltedRaddix:
Hi Sergey,
 
You are absolutely right on all three counts and I apologize for the noise. My previous analysis contained two claims that were incorrect:
 
1. Locales — You're correct, the Windows ANSI code page is used for open() across all locales on Windows, not just German. That was a sloppy description on my part and I should have been more precise by quoting the actual Python docs about ANSI CP behavior.
2. LZ-compressed config.txt — After re-reading change_to_lz_path() carefully and your code analysis, the "gcodes/" prefix guard is unambiguous: for Player::download_command("/sd/config.txt") the function constructs an invalid .lz sub-path, the first fwfs::fopen call returns NULL, and the firmware correctly falls back to the uncompressed /sd/config.txt. My 0x00 0x00 QuickLZ heuristic for config files was therefore unjustified — a classic case of me seeing binary bytes and jumping to the wrong root cause. Thank you for flagging this.
 
Regarding the AI style comment — fair enough, and I completely understand why the analysis came off as a wild goose chase. I do use AI assistance as a tool for phrasing, translation and code structure organization (it is 2026 after all), but the technical responsibility for what I submit is 100% mine, and in this case I submitted two incorrect technical claims without doing my due diligence on the firmware source first. That's on me.
 
Updated PR #732: I have completely rewritten it against current main (force-pushed). The fix is now stripped down to the absolute minimum 2-line change that is 100% unassailable:- carveracontroller/main.py finishLoadConfig() → open config file with encoding="utf-8", errors="replace"- carveracontroller/main.py load_machine_config_json() → open model JSON file with encoding="utf-8", errors="replace"
No LZ detection, no other code paths touched, no other files changed.
 
The bug itself is still real on Windows and independently reproducible: Python's open() on Windows without an explicit encoding argument defaults to the ANSI code page (cp1252 in practice), and cp1252 has five unmapped byte values (0x81, 0x8D, 0x8F, 0x90, 0x9D). If any file received from the machine over the framed protocol contains padding bytes, stray CRC bytes, cache junk, or literally any content that accidentally contains one of those five unmapped codepoints — f.read() throws "charmap codec can't decode byte ... character maps to <undefined>" and the controller enters the config-retry loop. On an English developer machine where the file contents happen to never collide with those 5 bytes, it would never trigger, which explains the "it works on my machine" asymmetry. Adding the explicit encoding with replace is the standard Python idiom for robustness here and it's what python -m py_compile and modern linters flag as a best-practice issue on cross-platform code.
 
 
Johannes (ExaltedRaddix)



Gesendet: Freitag, 21. August 2026 um 11:54
Von: SergeBakharev ***@***.***>
An: Carvera-Community/Carvera_Controller ***@***.***>
CC: ExaltedRaddix ***@***.***>,Mention ***@***.***>
Betreff: Re: [Carvera-Community/Carvera_Controller] Fix config load fail on QuickLZ-compressed downloads (ftype=lz) (PR #732)

SergeBakharev left a comment (Carvera-Community/Carvera_Controller#732)
Johannes,
The crash cannot be OS-Locale dependent. The reason I'm so confident that it's not because of your German locale is because Python's default filesystem encoding is always cp1252 on Windows irrespective of locale, per the Python documentation:

open() In text mode, if encoding is not specified the encoding used is platform-dependent: locale.getencoding() is called to get the current locale encoding.


locale.getencoding() ... On Windows, return the ANSI code page.

Please have a think about you might have done that is special. If you pull the SD card out of the machine and examine the creation date of the config file there might be a clue.
I looked at the player.cpp firmware code myself. It determines if a file is LZ compressed in an upload by the file extension suffix .lz:

//if file is lzCompress file,then need to Decompress
start_pos = filename.find(".lz");
string srcfilename=lzfilename;
string desfilename= filename;
if (start_pos != string::npos) {
	desfilename=filename.substr(0, start_pos);
	if(!decompress(srcfilename,desfilename,u32filesize,stream))
       {
           if (communication_protocol == PROTOCOL_MAKERA) { 
               sprintf(error_msg, "error: error in decompressing file!\r\n");	
           }
		goto upload_error;
       }
   }

We can't do the same in the Controller, since we are requesting the file /sd/config.txt, and per your claim receiving a .lz compressed stream instead.
Johannes, I've looked further at the firmware code and in fact there IS guardrails to prevent a compressed /sd/config.txt from being returned. Player::download_command always maps through change_to_lz_path(), which only knows about /sd/gcodes/.lz/:

// Change from origin path to quicklz file sub path
std::string change_to_lz_path( std::string origin )
{
	unsigned found = origin.find("gcodes/");
	string filename = origin.substr(found + 7);	
	string path = "/sd/gcodes/.lz";
	DIR * d = fwfs::opendir(path.c_str());
	if(NULL == d )
	{
		fwfs::mkdir(path.c_str(), 0);
	}
	else
	{
		closedir(d);	
	}
	return "/sd/gcodes/.lz/" + filename;
}

As an aside, I can't help but notice that your correspondence is written with AI assistance. While I have no issue with it's use for translation purposes, I can't help but feel that I've been put on a wild goose chase for a bug that doesn't exist.
I won't take this PR any further until you can show the code path that would lead to a LZ compressed config.txt being returned in response to the command download /sd/config.txt.

PR #733, opened 2026-08-12
Build: Windows pure-Python msgfmt fallback (no gettext tools needed)

ExaltedRaddix:
- Scripts that ship with the repo (build.py compile_mo step) called external msgfmt.exe unconditionally, which is not installed by default on Windows developer machines causing FileNotFoundError
- Prefer external msgfmt if present on PATH (existing behavior)
- Fall back to a minimal pure-Python .po -> .mo compiler covering the small message set used by carveracontroller (en + zh-CN), so builds work out-of-the-box without installing GNU gettext on Windows
- Tested: build.py --os windows completes without error on clean Windows

### DIFF
--- a/scripts/update_translations.py
+++ b/scripts/update_translations.py
@@ -85,11 +85,111 @@ def generate_po():
 
 def compile_mo():
     # Compile .po files to .mo files
+    import shutil as _shutil
+    import struct as _struct
     po_files = [f"{PACKAGE_PATH}/locales/{lang}/LC_MESSAGES/{lang}.po" for lang in LANGUAGES]
     for po_file in po_files:
         mo_file = po_file.replace(".po", ".mo")
-        subprocess.run(["msgfmt", "-o", mo_file, po_file], cwd=PACKAGE_PATH)
-        print(f"Compiled {po_file} to {mo_file}")
+        # Prefer external msgfmt if available; fall back to pure-Python compiler
+        msgfmt_exe = _shutil.which("msgfmt")
+        if msgfmt_exe is not None:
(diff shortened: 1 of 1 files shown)


SergeBakharev (changes requested):
msgfmt does plenty of things that this fallback python doesn't.

I think this is the case where it is more dangerous to pretend that the fallback is equivalent than to just skip update_translations. A developer working on translation support would have a potentially completely different result than the actual CI builds of the app. A simple example: msgfmt skips `#, fuzzy` by default. The fallback does not; it ships those msgstr values. zh-CN.po has many fuzzies (stale/guessed merges), e.g. " Offsets: " → "X偏移:". 

msgfmt only matters for shipping/updating non-English. For local development purposes, if you can read english there is no benefit.

Please replace the fallback with something like

```
def compile_mo():
    import shutil
    if not shutil.which("msgfmt"):
        print("msgfmt not found; skipping translation compile")
        return
```

and update the README.md

SergeBakharev:
@ExaltedRaddix Are you able to make the requested changes?

ExaltedRaddix:
Oh sorry for late reply. yes i think so. wait for my response ;)


Gesendet: Freitag, 21. August 2026 um 11:57
Von: SergeBakharev ***@***.***>
An: Carvera-Community/Carvera_Controller ***@***.***>
CC: ExaltedRaddix ***@***.***>,Mention ***@***.***>
Betreff: Re: [Carvera-Community/Carvera_Controller] Build: Windows pure-Python msgfmt fallback (no gettext tools needed) (PR #733)

SergeBakharev left a comment (Carvera-Community/Carvera_Controller#733)
@ExaltedRaddix Are you able to make the requested changes?

Issue #730, opened 2026-08-11
Allow jogging when machine running prevents smooth pendant movement


SergeBakharev:
The controller option "Allow Jogging When Machine Running" is set to default False.

This prevents jogging commands to be sent while the machine is already moving.

This is unintended. The purpose of this setting was to prevent jogging completely while the spindle was running.

For now I've written a note in the docs https://carvera-community.gitbook.io/docs/controller/features/pendant-support#jog-movement-buffer

But this safety feature should be fixed to actually look at the state of the spindle instead of the overall machine state.

SergeBakharev:
After some discussion it seems that while originally unintended, it is agreed as the safe default.

Will just revise the description:
> "This option allows for manual movement while spindle is running. Use this functionality with caution."

SergeBakharev:
or alternative suggestion split it into two different toggles. One preventing jogging while spindle is running, while the other prevents queuing multiple pendant jogs

PR #737, opened 2026-08-17
Add "Allow jogging when spindle or laser is on" setting

Lyrkan:
Related to https://github.com/Carvera-Community/Carvera_Controller/issues/730#issuecomment-5250466360

This adds a new "Allow jogging when spindle or laser is on" setting after the existing "Allow jogging when machine running":

image

By default the option is disabled on a new install, but takes the value of "Allow jogging when machine running" if it was already set (so same behaviors for existing users).

I did a bit of refactoring so that the current state can be accessed through `app.jog_controls_enabled` so we don't have to copy paste the same checks everywhere anymore (see `makera.kv` changes).

The current state of the spindle (**Edit:** and laser) is also now available in `app.spindle_or_laser_is_on`. It is considered enabled if:
* The current state is different from `NOT_CONNECTED` (for obvious reasons) and `CONNECTED` (which is in reality "Wait")
* And one of those two conditions is true:
  * The laser mode is disabled and `curspindle` is > 0 
  * The laser mode is enabled and `laserpower` is > 0

**EDIT:** I initially made it so that it only checked the spindle state (see first commit) and purposefully excluded the laser, but after thinking about it it doesn't make sense to allow jogging if the laser is on either, so I updated it in the second commit. Feel free to revert that one if you prefer to only check if the spindle is spinning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 [2.2.0-RC2]
+- Enhancement: Adds "Allow Jogging When Spindle Is On" option (disabled by default). "Allow Jogging When Machine Running" will now be enabled by default. Existing configs that already allowed jogging while the machine is running also enable the new spindle option.
 - Fixed: Prevent a probing modal crash if E is not provided when using the angle operation
 
 [2.2.0-RC1]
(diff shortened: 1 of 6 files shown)


SergeBakharev:
much cleaner, thank you!

PR #741, opened 2026-08-19
Moved Z1 Camera to a collapsible area in the Gcode Viewer

### LINKED ISSUE #716: Main menu button can trigger the camera resolution dropdown
A bit annoying to reproduce since the trigger area seems to be a pixel large and is not always placed at the same location...

Image

SergeBakharev:
Added a new UI element for us to use CollapsibleSplitter. It's like a normal Splitter but when clicked it collapsed/opens, and if dragged it resizes.

I've moved the Z1 CameraView to an area hidden by this CollapsibleSplitter, and tied toggling the CameraStream with the colapsed/open state. The Splitter doesn't show up if there is no supported Camera.

In future I would like to put my #587 into this collapsible area as well. Not sure yet how the toggle between the Camera view and Graph view should look (maybe two buttons? Both open = 50:50 split vertically?), but that's a problem for next release.

Not sure if it's the right stylist choice, but I intentionally had the Camera icon peeking when in the Collapsed state to help signal that it's hidden. @MichaelLevAstro @michael-ring  @Lyrkan Opinions?

Ultimately I'm doing this to resolve #716 

Screenshot 2026-08-19 at 11 22 31 pm

Screenshot 2026-08-19 at 11 22 44 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fixed: Z1 machine config backup option added to settings matching other models
 - Fixed: Switching between different machines no longer reuses the previous machine's settings panels or cached config.txt
 - Changed: added help button to probing screen confirmation/error popup for clarity
+- Changed: Moved Z1 Camera to a collapsible area in the Gcode Viewer. Collapsible splitter is only shown if a supported camera is found.
 
 [2.2.0-RC1]
 - Enhancement: Read tool definitions from post-processor outputs and use them in the G-code viewer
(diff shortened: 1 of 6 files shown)


Lyrkan:
> Not sure if it's the right stylist choice, but I intentionally had the Camera icon peeking when in the Collapsed state to help signal that it's hidden. @MichaelLevAstro @michael-ring @Lyrkan Opinions?

LGTM :+1:

MichaelLevAstro:
Looks nice!
The clipped cam button looks broken imo, i think having it fully visible is enough for user to understand he needs to click it in order to open the camera.

SergeBakharev:
With free eyes this morning, I agree it looks broken peeking like that. I've revised so that when collapsed it anchors to the bottom of the parent.

Screenshot 2026-08-20 at 9 36 37 am

Issue #692, opened 2026-07-22
Add support for Makera Z1


michael-ring:
Protocol of Z1 seems to be the same as Protocol for recent other Firmwares
Z1 in principle is already usable, for an initial port I have provided changes for Config Defaults and a new Config filwe

SergeBakharev:
Now that we have some initial support, @michael-ring what is remaining? Can you make a list?

I know there are a bunch of Z1 specific features shown on the config-and-run screen. We should probably hold off on adding those until we have them in the Community Firmware, but it doesn't stop us preparing.

Camera support is being worked on here: https://github.com/Carvera-Community/Carvera_Controller/pull/694/changes#top

michael-ring:
Here's what I see:

Firmware Update in Community Firmware needs some love, it currently offers the 2.1 firmware which is wrong for Z1 atm. But likely something not worth fixing because anyway Community Firmware is the goal

Camera support - Already in the works https://github.com/Carvera-Community/Carvera_Controller/pull/694

3D Probing - As discussed will not happen based on Makera Firmware

Display RSSI in Diagnostics Menu

Vacuum Management (Called auto Vacuum in Studio) & Vacuum Status in Diagnostics (Can work on this if you want me to)
Auto Blow
Auto Bed Clean
Antistatic --> This would be the only relevant thing missing (Can work on this if you want me to)

General Convenience feature:
Move to Vise (like Anchor 1 and Anchor 2) can make sense for the Z1 as there is a Makera Product. Idea would be to move to the front of the left jaw to make 3D probing a bit easier. We could reuse the Anchor2 configs if it is a nogo to create a specific setting

Jogging Support with Makera Firmware & Pendant is not working well, no idea if this will get smooth with Community Firmware or if this is a problem in Community Controller (Update Frequency??) or a MacOSX specific problem.

SergeBakharev:
> Firmware Update in Community Firmware needs some love, it currently offers the 2.1 firmware which is wrong for Z1 atm. But likely something not worth fixing because anyway Community Firmware is the goal

Agreed. Adding machine compatibility matrix is too much work. Once we add support for the Z1, I would expect it to remain.

> Display RSSI in Diagnostics Menu

This is an interesting point. We should add this in general, as it's available as of https://github.com/Carvera-Community/Carvera_Community_Firmware/commit/b48247b490a2dfd979df14cc81ffd4cfab31c6af Hopefully the Z1 implementation will expose it as well.

[...]

PR #754, opened 2026-08-23
Z1 usb support

SergeBakharev:
#692 

This adds Z1 USB support via userland implementation pyusb/libusb.

Linux users need a udev rule similar to:
`SUBSYSTEM=="usb", ATTR{idVendor}=="303a", ATTR{idProduct}=="4002", MODE="0660", GROUP="plugdev", TAG+="uaccess"`

MacOS users don't need anything if using the packaged version, developers just need to make sure the latest package deps from the lock file are installed.

Windows users are in the sucky position that the WinUSB driver needs to be installed, and since the firmware doesn't have the required Microsoft OS 2.0 descriptors for WinUSB, you need to install the driver via Ziadig. This is the same as the Makera software. We can do better once we have a Community firmware on the Z1.

I've got a popup in the UI if it's detected the USB VID/PID but missing WinUSB or the Linux udev rule to direct the user to https://carvera-community.gitbook.io/docs/controller/features/z1-usb-support I haven't created this page yet in docs.

One annoyance i've found is that the Z1's usb serial string is set to `123456`. This isn't some placeholder in the Controller, that's actually what Makera have in their Z1 firmware.

I've tested MacOS support of this PR, but I haven't touched Linux or Windows yet


SergeBakharev:
closing the app while connected the Z1 over usb causes it to crash because of some GC issue. This needs further revision

PR #759, opened 2026-08-25
Probing: Make Y+/Y- labels respect the inverted Y-axis jogging setting

### LINKED ISSUE #757: The Y-Jogging in Probing Wizzard does not follow the setting "Invert Y Axis Jogging Buttons and Keyboard Controls"
Easy to test...

turn on 

"Invert Y Axis Jogging Buttons and Keyboard Controls" 

You will see that on main page the keyboard and the buttons change.

Now go to 'Probing'

There the buttons and the keyboard now work as if the Setting were off.

One other small detail, the settings from the main controls are not taken over. So when you jogged to position and used '1' as step size then in the Probing Dialog you will be surprised when you want to do another small adjustment because default there is 10

acj:
This is two related fixes for jogging. The step size sync happens in both directions. I'm not certain that we want changes to the step size on the Probing screen to propagate back to the main screen, but it seemed intuitive in my testing and is less to keep track of.

Fixes #757


SergeBakharev:
I agree that propagating them back to the main screen seems right. PR looks good, just re-targetting it to release/2.2.x for now (not sure if we will do a RC3) and I'll cascade it down into `develop`. Either way it will end up in `develop`

SergeBakharev (approved):

PR #760, opened 2026-08-26
Implement web-based monitoring with time estimate and 3D gcode toolpath on port 8765

waynepiekarski:
I have a laptop with pendant attached to the Carvera Air that I use to do probing and then start the job. But then I want to track the progress from my phone or my work laptop to see when the job will be finished. This change adds a web server to the controller on port 8765 that you can connect to via a web browser on the same network. It shows a live time estimate, 3D rendering of the gcode toolpath, and the position of the tool. I put this prototype together with Claude Code to see what could be possible and I was happy with the result. The alternative would be to have a read-only controller that somehow connects to the existing controller and proxy the connection through, but that seemed too difficult.

### DIFF
--- a/carveracontroller/controller_config.json
+++ b/carveracontroller/controller_config.json
@@ -322,5 +322,21 @@
         "section": "carvera",
         "key": "show_playbar_tool_change_markers",
         "default": "true"
+    },
+    {
+        "type": "bool",
+        "title": "Enable Job Status Web Server",
+        "desc": "Serves a read-only web page showing job progress (elapsed/remaining time, % complete) on your local network, so another device can watch it. This opens a network port, so it is off by default. Restart the controller after changing this.",
+        "section": "carvera",
+        "key": "status_server_enabled",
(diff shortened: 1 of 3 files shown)


SergeBakharev:
There is a somewhat similar PR to add this functionality to the machine firmware: https://github.com/Carvera-Community/Carvera_Community_Firmware/pull/391 

I personally prefer the machine based solution as it's more self contained. However there is a discussion about whether or not the ESP8266 inside the machine is up to task for multiple connections, last thing we want is to have the machine crashing.

On the flip side the time remaining estimate on the Controller is a lot more accurate since we estimate it based on actual gcode movement/distance/speed etc.

waynepiekarski:
Yeah I wrote this into the controller because I didn't want to burden the firmware with extra work that it might not be able to handle, and seemed easy enough to get it to work since all the information was available to do it.

So can we merge this or do you have concerns about the implementation?

PR #770, opened 2026-08-29
Time until toolchange details

SergeBakharev:
Using the Z1 has made me want a bit of information that isn't as relevant for the ATC machine; when will the next tool change occur.

This PR adds this information in two ways:
1. As a tooltip on the Toolchange flags
Screenshot 2026-08-29 at 11 35 32 am

2. In the progressbar itself, alternating between time remaining for the entire program and until the next tool change. Ignore the difference in time here in the screenshot, I pushed play but then hovered...

Screenshot 2026-08-29 at 11 39 33 am

In adding this I noticed that a time estimate wasn't shown when a file had been selected but not running, so I've added that now too:
Screenshot 2026-08-29 at 11 41 34 am

@Lyrkan Tell me what you think, I think the design language is a bit funny to have tooltips on the tool flags but I can't imagine a better way

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 [unreleased]
+- Enhancement: Tool-change flags have tooltip showing time until the change
+- Enhancement: Remaining time text alternates with time until the next tool change and playback completion
+- Enhancement: Selected files show estimated run time on the playback bar before the job starts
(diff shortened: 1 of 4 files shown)


Lyrkan:
Seems good to me, I like the idea.
My only nitpick would be the vertical alignment of the text in the tooltip, but that's really not a big issue.

SergeBakharev:
Good eye!

PR #778, opened 2026-08-31
Fix QuickLZ decoding for downloaded files

ExaltedRaddix:
bugfix/config-quicklz-download-decode

### DIFF
--- a/carveracontroller/XMODEM.py
+++ b/carveracontroller/XMODEM.py
@@ -628,6 +628,7 @@ class XMODEM:
         expected_md5 = None
         received_md5 = hashlib.md5()
         first_bytes = bytearray()
+        packet_size = 0
         self.deferred_download_md5 = None
         self.download_md5_failed = False
         self.FileRcvState = FileTransState.WAIT_MD5
@@ -659,10 +660,18 @@ class XMODEM:
                             data_len = ((self.packetData[0] << 8) | self.packetData[1]) - 7
                             income_size += data_len
                             payload = self.packetData[7 : (data_len + 7)]
+                            self.log.debug(
+                                "PTYPE_FILE_DATA seq=%d/%d payload=%d bytes expected-max=%d first=%s",
+                                seq,
+                                total_packet,
+                                len(payload),
+                                total_packet * packet_size,
+                                bytes(payload[:32]).hex(" "),
+                            )
                             stream.write(payload)
                             received_md5.update(payload)
-                            if len(first_bytes) < 2:
-                                first_bytes.extend(payload[: 2 - len(first_bytes)])
+                            if len(first_bytes) < 32:
+                                first_bytes.extend(payload[: 32 - len(first_bytes)])
                             if sequence < total_packet:
                                 sequence += 1
                                 data = sequence.to_bytes(4, byteorder="big", signed=False)
@@ -697,6 +706,13 @@ class XMODEM:
                                 | (self.packetData[5] << 8)
                                 | self.packetData[6]
                             )
+                            packet_size = (self.packetData[7] << 8) | self.packetData[8]
+                            self.log.debug(
+                                "PTYPE_FILE_VIEW packets=%d packet-size=%d expected-max=%d",
+                                total_packet,
+                                packet_size,
+                                total_packet * packet_size,
+                            )
                             sequence = 1
                             data = sequence.to_bytes(4, byteorder="big", signed=False)
                             self._send_file_trans_command(PTYPE_FILE_DATA, data)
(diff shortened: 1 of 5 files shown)


ExaltedRaddix:
## Summary

- Decode QuickLZ-compressed download payloads before loading or saving the requested file.
- Validate the QuickLZ block structure, checksum, decoded size, and advertised MD5.
- Preserve explicitly requested `.lz` files.
- Add diagnostic logging and tests for framed Makera and legacy XMODEM transfers.

## Testing

- 33 unit tests passed.
- Successfully tested on a machine running Carvera Community Firmware 2.2.0c-RC3
- `/sd/config.txt` now loads correctly without ignoring or replacing invalid bytes.

ExaltedRaddix:
Hey @SergeBakharev @acj — friendly ping on this one 🙂 Been running the fix on Carvera Community Firmware 2.2.0c-RC3 for about a week, `/sd/config.txt` loads correctly now. No rush, just checking if there's anything I should adjust for review. Thanks!

SergeBakharev:
Johannes, my opinion hasn't changed since [my last comment](https://github.com/Carvera-Community/Carvera_Controller/pull/732#issuecomment-5368337733) to your previous PR. I don't believe detecting if config.txt is LZ compressed to be an actual issue, and until you can show me how a LZ compressed config.txt could be returned in response to `download /sd/config.txt` any additional LZ detection code is unnecessary additional complexity.

If you think there is a bug in the firmware that allows for a non-ascii character to be saved to config.txt via `config-set`, or an issue with the MD5 hash checking of transferred files, I'll be receptive to targeted fix. But first please open an issue describing the problem instead of burning LLM tokens.

ExaltedRaddix:
## Reproducible proof:  download returns QuickLZ binary

@SergeBakharev Here is concrete evidence that the problem is real. I downloaded  from my Carvera running 2.2.0c-RC3 today with the current controller, and **the file is a QuickLZ-compressed binary blob**, not plain text.

### The downloaded 

- **File size:** 2692 bytes
- **NOT valid UTF-8:** opening it with  (without ) crashes with  on Windows
- **NOT valid ASCII:** contains bytes like , 

### Structured proof it's a QuickLZ container



### Human-readable text fragments visible in the compressed payload



You can verify this yourself with any hex editor — the file at offset 4 is a perfectly valid QuickLZ level-1 header.

### Why this happens

The firmware caches config.txt internally as a QuickLZ container on the SD card (this is how the firmware stores it). Under certain conditions (possibly after a config-set via , or after the firmware reads it from cache during a concurrent operation), the download endpoint returns the **cached QuickLZ blob** instead of the decoded plain text.

This is **not a hypothetical** — it just happened on my machine 10 minutes ago with a fresh download.

### This PR fixes the real problem

The current controller has **zero defense** against this. When the firmware sends a LZ blob for , the controller:

1. Writes the binary blob to disk as 
2. Tries to open it as text ?  crash on Windows
3. Even on Linux, ConfigParser fails on the binary content

This PR adds a single call to  that:

- Detects the QuickLZ container structure (not heuristic — validates the actual block format and checksum)
- Decompresses it to the actual config text
- Verifies the MD5 matches what the firmware advertised
- Passes through plain-text downloads unchanged (zero overhead for the normal case)

The controller should be resilient regardless of what the firmware sends. Just like a web browser handles malformed HTML instead of crashing.

ExaltedRaddix:
## Screenshot: the crash in the current controller

This is a fresh download from 10 minutes ago with the **current main-branch controller** connected to a Carvera running 2.2.0c-RC3:



This is exactly the crash path this PR fixes. The firmware returns a QuickLZ-compressed binary blob, and the controller tries to open it as text without encoding handling, choking on byte 0x81.

Byte 0x81 is not valid in any Windows code page. It is part of the QuickLZ stream. The file at offset 51 corresponds to the second half of the first QuickLZ block compressed stream.

### This is not just a firmware bug

Even if the firmware gets patched tomorrow to never return LZ for config downloads, **every Carvera in the field running 2.2.0c-RC3 will still crash** until either the firmware is updated (which end users rarely do) or the controller handles it.

Defense in depth. The controller must not crash because the firmware sent unexpected bytes.

ExaltedRaddix:
[config.txt](https://github.com/user-attachments/files/32056569/config.txt)
photo_2026-09-10_13-38-00

SergeBakharev:
@ExaltedRaddix I was able to reproduce the issue with the following steps:
1. Using the Controller upload a config.txt to the highest level directory selectable in the Controller. This uploads config.txt to the `gcodes/` directory, not the root of the SD where it needs to be to be actually used by the machine.
2. Close and reopen the controller
3. On reopen the controller tries to `download /sd/config.txt` and instead gets the LZ compressed file from `/sd/gcodes/.lz/`

This is a firmware bug. Even if the controller was to decompress the file correctly, the file isn't representative of what is actually used as the config.txt by the machine, so this PR wouldn't resolve the actual issue. 

Tracking the firmware bug as https://github.com/Carvera-Community/Carvera_Community_Firmware/issues/440

On your machine you can resolve this issue by using the Controller to delete `/gcodes/config.txt` in the file browser, this will also delete the /sd/gcodes/.lz/config.txt file causing the issue.

PR #789, opened 2026-09-04
File browser UI improvements

Lyrkan:
This PR is an attempt at resolving #509 

In reality it doesn't change much in terms of process compared to what we already have, but instead of nesting multiple popups it brings everything into a single tabbed one. It also affects the "upload" part of the firmware update popup: same thing but without the tabs, without most of the actions, and with a filter on .bin files.

Additionally, I added a way to display the preview images embedded by Makera Studio when they are available. It shouldn't be that heavy since it only has to check the end of gcode files and keep track of whether or not they have an image (both for local and machine files, for the latter this process happens when the file is downloaded). If they have one it is extracted into a cache folder in the app directory (with only the last 200 being kept).

Last thing: I noticed a bug with the config backup download (files were applied / opened in the gcode viewer) and fixed it on this branch , but it seemed that it already existed on `develop`. I left it here since I also modified `doDownload` for this feature.

<br/>
<br/>

As usual, some screenshots:

<br/>

**While disconnected (local device tab):**
image

<br/>
<br/>

**While disconnected (machine tab):**
image

<br/>
<br/>

**G-Code image preview on hover:**
image

<br/>
<br/>

**Connected with a file being previewed:**
image

<br/>
<br/>

**Multi select (only supports file removal for now):**
image

<br/>
<br/>

**Connected (machine tab):**
image

<br/>
<br/>

**Firmware upload:**
image


Lyrkan:
~Hm, looks like it's crashing on my Android tablet for some reason, switching it to draft~
**Edit: all good now, just had to move Pillow to the runtime dependencies**

Also I forgot to mention it but I've got some mild concerns about iOS: as far as I know it doesn't allow the kind of listing we do in the "This device" tab so it should instead display a message and a browse button that opens the native filepicker. It should then work as before but since I don't own any Apple devices I cannot test it.

SergeBakharev:
Don't worry about iOS, I've got some hardware coming so I can provide on going support there.

I like how we can now download files with the browser now! Overall it looks more modern, and easier to understand.

I've made two small nitpicky changes:
* "Your Machine" became "CNC Machine"
* When a file was selected, in the browser it would have a badge that read "Current", I changed this to "Selected"

Screenshot 2026-09-09 at 10 27 57 pm

Ideally I think we should also change the Makera logo to a milling machine, but I can't find a suitable Material Design style icon.

This upper area seems a bit wasteful, but I don't have any suggestions for how to do it better. Moving it to the bottom will make it confusing about which file the action buttons are acting on.

Screenshot 2026-09-09 at 10 35 00 pm

For the **Recent Places** button, I don't think a folder icon is the right picture to convey it's purpose. Maybe a [bookmark](https://pictogrammers.com/library/mdi/icon/bookmark-box-multiple/) would be better?

Lyrkan:
> I've made two small nitpicky changes:
> 
>   * "Your Machine" became "CNC Machine"
> 
>   * When a file was selected, in the browser it would have a badge that read "Current", I changed this to "Selected"

👍 

> Ideally I think we should also change the Makera logo to a milling machine, but I can't find a suitable Material Design style icon.

Yeah, I looked at the entire list multiple times but couldn't find anything either.

> This upper area seems a bit wasteful, but I don't have any suggestions for how to do it better. Moving it to the bottom will make it confusing about which file the action buttons are acting on.

I also tried moving that line elsewhere (below the tabs, at the bottom, directly in the title bar) but didn't find a better place and ended up getting used to it being there.

We could remove it since the current filename is already displayed outside of that modal but that would mean either:
* also removing the "clear" button with no way of doing this action anymore (I have no idea if/how much it is actually used)
* leaving it with no explanation of what it does and no feedback when it is pressed...

> For the **Recent Places** button, I don't think a folder icon is the right picture to convey it's purpose. Maybe a [bookmark](https://pictogrammers.com/library/mdi/icon/bookmark-box-multiple/) would be better?

Hmm, the folder icon kind of made sense to me, but now that you're saying this I agree that maybe it could be improved.
I don't know about the bookmark one since users cannot directly define what appear in the list. Another option would be [history](https://pictogrammers.com/library/mdi/icon/history/).

SergeBakharev:
Oh nice, the history icon is better than the bookmark icon. Let's use that one.

The rest let's leave as-is, I don't see any way to do better for now. Maybe if we ask nicely, and Fae has time she could draw us a CNC icon.

PR #791, opened 2026-09-04
Fix 4-axis viewer transform order to rotate around the WCS origin before centering

Lyrkan:
Fixes an issue with how the 4 axis paths are rendered when they do not wrap evenly around the X axis (which is why we didn't notice it on files like Nefertiti).

This was caused by a wrong application order of some transforms:
* Before: world position -> center translation -> rotary rotation -> camera/view -> projection
* After: world position -> rotary rotation -> center translation -> camera/view -> projection

<br/>

For instance, using the following code:
```
(@F360|STOCK|id=cylinder|width=20|depth=20|height=20|diameter=20)
(@F360|ORIGIN|type_name=leftCenter|x=-10|y=0|z=0)
M321
G0Z10
G00 G17 G21 G54
G90
G0 X0 Y0 Z10 A0
M3
G1 A60 S0.6 F300
G1 A120
G1 A180
G1 S0
M5
M2
```

<br/>
<br/>

**Before:**
image

<br/>
<br/>

**After:**
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fixed: Harden the gcode parser against "zero length" movement, and prevent division by zero in play slider
 - Fixed: Time estimates ignoring speed for some 4th-axis moves
 - Fixed: Allow to select the bottom element of the MDI, Gcode and probing confirmation lists
+- Fixed: Fixed rotary previews so toolpaths, pointers, and stock rotate around the WCS origin without unwanted orbiting.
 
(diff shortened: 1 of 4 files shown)


Lyrkan:
This also fixes another bug I noticed tonight while looking at a file sent by someone on Discord: the tool pointer was sometimes jumping in and out for no reason at all.

Before:
bug-4th-axis-tool-jump-01

After:
bug-4th-axis-tool-jump-02

PR #799, opened 2026-09-10
Improve update popup UI and retrieve version data from the GitHub API

Lyrkan:
# Summary

This PR replace the current firmware/controller update popup by a new one with a better UI and based on GitHub releases (instead of reading the content of changelogs, see #652).

Recording 2026-09-04 at 19 41 43

<br/>
<br/>

Some notes:
* the popup allows to select either the latest stable release or latest RC
* the release notes are directly retrieved from the release data of the selected version (it's just a formatted version of the markdown content)
* since we are calling the GitHub API (and also doing it anonymously) the release data is put into a cache file (in the app directory) to avoid encountering quota issues
* the "Install Firmware x.y.z" button does an (almost) one-click install (download + checksum verification + copy on the machine + ask for reset). For now I didn't make it available for the Z1/Z1PRO since I'm not sure how this process goes and have no idea how the asset will be called in the release (but I did have to make some assumptions about the fact that its name will contain "z1" in the `select_firmware_asset` method so it won't break this feature once we release it).

I tested this on Linux with my CA1 and used Cursor with various models when developing it.

# Contributor checklist

- [x] I am a human submitting this pull request.
- [x] I have read and understand every changed line, and I take responsibility
      for the complete change.
- [x] Pre-commit checks (`poetry run pre-commit run --all-files`) passes with no errors.
- [x] The changes have been tested with real hardware.
- [x] I have read and understood the [contributions guidelines](https://github.com/Carvera-Community/Carvera_Controller/blob/develop/CONTRIBUTING.md)


SergeBakharev:
Looks great, maybe we should add a way of selecting the "dev" release? Though this might result in novice users using it without a good way of providing us feedback...
Something to think about.

You are right to not add Z1 support for firmware binaries, we need to flesh out if we will have a single binary with both LPC+ESP firmware, just LPC, or have them separately upgradable.

PR #821, opened 2026-09-19
Stop splitting 3-axis G0/G1 moves every 0.5mm and split A-axis moves using the same tolerance as G2/G3 instead of every 0.5°

Lyrkan:
# Summary

The parser currently split all G0/G1 moves every 0.5mm in order to support A axis moves (so they are split every 0.5°) which can be really inefficient.

This PR:
* Stops splitting 3-axis-only moves
* Replaces the current "every 0.5°" split for A-axis moves by the principle used for G2/G3 arc moves (based on `CNC.accuracy` which allows a  0.01mm error)

When applied to some `carveracontroller/gcodes/Examples` files:

| File                        | Vertices before | Vertices after |
|-----------------------------|-----------------|----------------|
| LED/ABS-Base.nc             | 61,935          | 6,851          |
| LED/ACRYLIC-Balloon.nc      | 21,189          | 3,433          |
| LED/ACRYLIC-Carvera.nc      | 45,507          | 21,557         |
| Laser/AudreyHepburn.nc      | 320,156         | 269,258        |
| Laser/AudreyHepburnSmall.nc | 37,441          | 31,539         |
| Relief/PirateShip.nc        | 249,180         | 189,616        |
| Rotation/NefertitiFinish.nc | 289,317         | 103,595        |
| Rotation/NefertitiRough.nc  | 317,075         | 40,209         |

<br/>
<br/>

image

image

image

<br/>
<br/>

I tested this on Linux and used Cursor with various models when developing it.

# Contributor checklist

- [x] I am a human submitting this pull request.
- [x] I have read and understand every changed line, and I take responsibility
      for the complete change.
- [x] Pre-commit checks (`poetry run pre-commit run --all-files`) passes with no errors.
- [x] The changes have been tested with real hardware.
- [x] I have read and understood the [contributions guidelines](https://github.com/Carvera-Community/Carvera_Controller/blob/develop/CONTRIBUTING.md)

PR #824, opened 2026-09-22
Stock simulation: Fix carving only happening at the end of a move and improve heightmap carver shading

Lyrkan:
# Summary

This PR improves two things related to the stock simulation :

* The stock was only being cut at the end of each segment. This wasn't a big issue before #821 since G0/G1 segments were cut every 0.5mm but was a lot more visible after it on long moves. Now it should sync the cut with the current tool location.
* The shading used for the heightmap carver was a bit disappointing since all normals were either pointing directly up/down or toward a side. For instance, looking at an engraving from above only displayed a single color (except for the "default" material which was kind of a hack that darkened the color based on the depth). The new method should shade the heightmap slopes and also darken the lower stock based on height for other materials.

<br/>
<br/>

**Some before/after for the shading part:**

stock-shading-01-before

stock-shading-01-after

<br/>
<br/>

stock-shading-02-before

stock-shading-02-after

<br/>
<br/>

stock-shading-03-before

stock-shading-03-after

<br/>
<br/>

I tested this on Linux and used Cursor with various models when developing it.

# Contributor checklist

- [x] I am a human submitting this pull request.
- [x] I have read and understand every changed line, and I take responsibility
      for the complete change.
- [x] Pre-commit checks (`poetry run pre-commit run --all-files`) passes with no errors.
- [x] The changes have been tested with real hardware.
- [x] I have read and understood the [contributions guidelines](https://github.com/Carvera-Community/Carvera_Controller/blob/develop/CONTRIBUTING.md)


SergeBakharev:
Looks great!

